PR #14, opened 2015-02-17
V1.2.4 release candidate

brentpicasso:
This is a combo PR that includes:

- improvements to android Bluetooth processing and android behavior

- introduction of filters into sample stream. Filters can modify / add to the sample stream on the fly. in this case, adding the channel 'BestLap' and 'LapDelta' needed for the dashboard pages.

- Misc fixes.

Sorry this is a combo PR. I'll go through and annotate the changes.

### DIFF
--- a/autosportlabs/comms/androidcomms.py
+++ b/autosportlabs/comms/androidcomms.py
@@ -99,6 +99,10 @@ class AndroidComms(object):
     def keep_alive(self):
         self.send_service_command(SERVICE_CMD_KEEP_ALIVE)
     
+    def cleanup(self):
+        print('comms.cleanup()')
+        self.send_service_command(SERVICE_CMD_EXIT)
+        
     def close(self):
         print('comms.close()')
         if self.isOpen():
--- a/autosportlabs/racecapture/databus/filter/bestlapfilter.py
+++ b/autosportlabs/racecapture/databus/filter/bestlapfilter.py
@@ -1,11 +1,12 @@
-BEST_LAPTIME_KEY = 'BestLap'
+
 
 class BestLapFilter(object):
     """Update BestLap if laptime is present and is faster than the current best"""
+    BEST_LAPTIME_KEY = 'BestLap'
     best_laptime = 0
     best_laptime_meta = None
-    def __init__(self, channel_metas):
-        self.best_laptime_meta = channel_metas.get_meta(BEST_LAPTIME_KEY)
+    def __init__(self, system_channels):
+        self.best_laptime_meta = system_channels.findChannelMeta(self.BEST_LAPTIME_KEY)
         
     def get_channel_meta(self):
         return self.best_laptime_meta
@@ -19,5 +20,5 @@ class BestLapFilter(object):
             current_best_laptime = self.best_laptime
             if current_best_laptime == 0 or laptime < current_best_laptime: 
                 current_best_laptime = laptime
-                channel_data[BEST_LAPTIME_KEY] = current_best_laptime
+                channel_data[self.BEST_LAPTIME_KEY] = current_best_laptime
                 self.best_laptime = current_best_laptime
--- a/autosportlabs/racecapture/databus/filter/bestlapfilter.py
+++ b/autosportlabs/racecapture/databus/filter/bestlapfilter.py
@@ -1,0 +1,17 @@
+BEST_LAPTIME_KEY = 'BestLap'
+
+class BestLapFilter(object):
+    best_laptime = 0
+    best_laptime_meta = None
+    def __init__(self, value, channel_metas):
+        self.best_laptime_meta = channel_metas.get_meta(BEST_LAPTIME_KEY)
+        
+    def get_channel_meta(self):
+        return self.best_laptime_meta
+        
+    def filter(self, channel_data):
+        laptime = channel_data.get('LapTime')
+        if laptime != None:
+            if laptime < self.best_laptime: 
+                self.best_laptime = laptime
+            channel_data[BEST_LAPTIME_KEY] = self.best_laptime
--- a/autosportlabs/racecapture/databus/filter/bestlapfilter.py
+++ b/autosportlabs/racecapture/databus/filter/bestlapfilter.py
@@ -1,6 +1,7 @@
 BEST_LAPTIME_KEY = 'BestLap'
 
 class BestLapFilter(object):
+    """Update BestLap if laptime is present and is faster than the current best"""
     best_laptime = 0
     best_laptime_meta = None
     def __init__(self, value, channel_metas):
@@ -8,10 +9,15 @@ class BestLapFilter(object):
         
     def get_channel_meta(self):
         return self.best_laptime_meta
-        
+    
+    def reset(self):
+        self.best_laptime = 0
+         
     def filter(self, channel_data):
         laptime = channel_data.get('LapTime')
-        if laptime != None:
-            if laptime < self.best_laptime: 
-                self.best_laptime = laptime
-            channel_data[BEST_LAPTIME_KEY] = self.best_laptime
+        if laptime != None and laptime > 0:
+            current_best_laptime = self.best_laptime
+            if current_best_laptime == 0 or laptime < current_best_laptime: 
+                current_best_laptime = laptime
+                channel_data[BEST_LAPTIME_KEY] = current_best_laptime
+                self.best_laptime = current_best_laptime
--- a/autosportlabs/racecapture/databus/filter/bestlapfilter.py
+++ b/autosportlabs/racecapture/databus/filter/bestlapfilter.py
@@ -4,7 +4,7 @@ class BestLapFilter(object):
     """Update BestLap if laptime is present and is faster than the current best"""
     best_laptime = 0
     best_laptime_meta = None
-    def __init__(self, value, channel_metas):
+    def __init__(self, channel_metas):
         self.best_laptime_meta = channel_metas.get_meta(BEST_LAPTIME_KEY)
         
     def get_channel_meta(self):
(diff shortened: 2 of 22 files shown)


autosportlabs/comms/androidcomms.py
```diff
@@ -99,6 +99,10 @@ def open(self):
     def keep_alive(self):
         self.send_service_command(SERVICE_CMD_KEEP_ALIVE)
     
+    def cleanup(self):
+        print('comms.cleanup()')
+        self.send_service_command(SERVICE_CMD_EXIT)
+        
     def close(self):
```
brentpicasso: adding the concept of comms component to 'clean up' . In this case, a signal to the android service to terminate.

autosportlabs/racecapture/databus/filter/bestlapfilter.py
```diff
@@ -0,0 +1,24 @@
+
+
+class BestLapFilter(object):
+    """Update BestLap if laptime is present and is faster than the current best"""
+    BEST_LAPTIME_KEY = 'BestLap'
```
brentpicasso: This is a sample filter that maintains the channel 'BestLap' as data streams in from RCP.

autosportlabs/racecapture/databus/filter/laptimedeltafilter.py
```diff
@@ -0,0 +1,34 @@
+
+class LaptimeDeltaFilter(object):
+    """Update the DeltaTime if BestLap and a reference lap time is available.
```
brentpicasso: This is a sample filter that maintains the channel 'LapDelta' based on comparing best lap against PredTime or LapTime, depending on which is available.

autosportlabs/racecapture/settings/prefs.py
```diff
@@ -118,7 +118,7 @@ def load(self):
                     for id, channel in content_dict["gauge_settings"].iteritems():
                         self._prefs_dict["gauge_settings"][id] = channel
 
-        except IOError:
+        except Exception:
             pass
```
brentpicasso: Made this catch broader. We had cases where an empty file was being saved, triggering a JSON exception. figured we'd catch and re-save a good file the next time around.

autosportlabs/racecapture/views/dashboard/widgets/timedelta.py
```diff
@@ -33,7 +33,7 @@ def on_value(self, instance, value):
             railedValue = value
             railedValue = MIN_TIME_DELTA if railedValue < MIN_TIME_DELTA else railedValue
             railedvalue = MAX_TIME_DELTA if railedValue > MAX_TIME_DELTA else railedValue
-            self.valueView.text = '{0:+1.1f}'.format(float(railedValue))
+            self.valueView.text = u'{0:+1.1f\u2206}'.format(float(railedValue))
         self.update_delta_color()
```
brentpicasso: Adding the little triangle symbol next to the time delta view on the dashboard screens.

buildozer.spec
```diff
@@ -56,7 +56,7 @@ fullscreen = 1
 #
 
 # (list) Permissions
-android.permissions = INTERNET, BLUETOOTH, BLUETOOTH_ADMIN , WRITE_EXTERNAL_STORAGE
+android.permissions = INTERNET, BLUETOOTH, BLUETOOTH_ADMIN , WRITE_EXTERNAL_STORAGE, WAKE_LOCK
```
brentpicasso: Prevent android from sleeping when app is running.

mappedspinner.py
```diff
@@ -26,7 +26,7 @@ def setValueMap(self, valueMap, defaultValue):
         self.values = values
                     
     def setFromValue(self, value):
-        self.text = self.valueMappings.get(value, self.defaultValue)
+        self.text = str(self.valueMappings.get(value, self.defaultValue))
```
brentpicasso: A UI bug fix.

resource/channel_meta/system_channels.json
```diff
@@ -406,6 +406,24 @@
...
+            "nm": "LapDelta",
+            "ut": "Min",
+            "sys": 1,
+            "type": 7,
+            "prec": 1,
+            "min": 0,
+            "max": 0
+        },
+        {
             "nm": "LapCount",
```
brentpicasso: Adding some new channels in the system Channels table.

service/autosportlabs/comms/bluetooth/bluetoothconnection.py
```diff
@@ -33,58 +25,41 @@ def get_available_ports(self):
...
+                error_message = None
+            else:
+                error_message = 'Could not detect device {}'.format(port)
+        except Exception as e:
+            print('error opening socket ' + str(e))
+            error_message = 'Error opening Bluetooth socket: {}'.format(str(e))
+
+        if error_message != None:
+            raise Exception(error_message)
         if self.socket == None:
```
brentpicasso: greatly simplified the opening of a BT connection, now that we're running in an external service.

service/main.py
```diff
@@ -1,8 +1,11 @@
 from time import sleep, time
```
brentpicasso: Cleaned up background service, added the jnius.detach() call that the kivy maintainers just added to prevent service from segfaulting when exiting.

Jeff-Ciesielski: Just a thought, would it be better to encapsulate this service in a class? If we have additional services in the future (socket for example) that could help with interop.  Also, it would allow us to avoid the use of globals.

brentpicasso: I was thinking of that too.

On Tue, Feb 17, 2015 at 3:31 PM, Jeff Ciesielski <notifications@github.com>
wrote:

> In service/main.py
> <https://github.com/autosportlabs/RaceCapture_App/pull/14#discussion_r24865991>
> :
>
> > @@ -1,8 +1,11 @@
> >  from time import sleep, time
>
> Just a thought, would it be better to encapsulate this service in a class?
> If we have additional services in the future (socket for example) that
> could help with interop. Also, it would allow us to avoid the use of
> globals.
>
> —
> <https://github.com/autosportlabs/RaceCapture_App/pull/14/files#r24865991>
> .
>



-- 
Brent Picasso

Jeff-Ciesielski:
Overall, it's a bit broad, but none of the changes appear to be too scary.  My comment RE: Encapsulating BT service in a class can wait till later.  :+1:

brentpicasso:
I think the class refactoring on the service would be good for later and
possibly include the datastore as well; maybe as a 2nd service.

Sorry there are so many changes!

On Tue, Feb 17, 2015 at 3:44 PM, Jeff Ciesielski <notifications@github.com>
wrote:

> Overall, it's a bit broad, but none of the changes appear to be too scary.
> My comment RE: Encapsulating BT service in a class can wait till later. [image:
> :+1:]
>
> —
> <https://github.com/autosportlabs/RaceCapture_App/pull/14#issuecomment-74780724>
> .
>



-- 
Brent Picasso

autosportlabs/racecapture/databus/filter/bestlapfilter.py
```diff
@@ -0,0 +1,24 @@
...
+    def get_channel_meta(self):
+        return self.best_laptime_meta
+    
+    def reset(self):
+        self.best_laptime = 0
+         
+    def filter(self, channel_data):
+        laptime = channel_data.get('LapTime')
+        if laptime != None and laptime > 0:
+            current_best_laptime = self.best_laptime
```
ryandoherty: Why the reassignment here? Seems like it could be a little simpler unless there's something I missed.

service/main.py
```diff
@@ -30,67 +33,73 @@ def tx_api_callback(message, *args):
 def cmd_api_callback(message, *args):
     message = message[2]
     if message == 'EXIT':
+        Logger.info('RC_SVC: ############################## GOT EXIT CMD #####################################')
```
ryandoherty: Why Logger instead of print like everything else?

ryandoherty:
Most things seem sane to me. For future PRs we should separate them out into a separate branch per feature that is then made a PR. 

Yay for tests! :+1:

autosportlabs/racecapture/databus/filter/bestlapfilter.py
brentpicasso: where is the re-assignment?

service/main.py
brentpicasso: finally using the built-in Kivy Logger object like I should've been doing all along

autosportlabs/racecapture/databus/filter/bestlapfilter.py
ryandoherty: `current_best_laptime = self.best_laptime`

Seems like you could just use self.best_laptime?

service/main.py
ryandoherty: :+1:

autosportlabs/racecapture/databus/filter/bestlapfilter.py
brentpicasso: this is to prevent repeated referencing the member variable via self, by putting the member variable on the local stack, using it, and then putting it back when done. The latest version of the code shows this better, through line 24;

https://github.com/autosportlabs/RaceCapture_App/blob/v1.2.4_release_candidate/autosportlabs/racecapture/databus/filter/bestlapfilter.py#L24

ryandoherty: Yes, but why? :) What is wrong with referencing the value via `self`? Are we gaining speed, flexibility, something I'm not seeing?

brentpicasso: In most programming languages it's considered a best practice b/c faster and also so that another thread doesn't change the member variable out from underneath you (makes the work a more atomic). I wasn't sure if python local variables are faster than members so I looked; seems like they are: http://stackoverflow.com/questions/12397984/why-is-local-variable-access-faster-than-class-member-access-in-python

ryandoherty: Yeah, accessing a local var is usually faster in all languages. Is this code multithreaded? 

Seems like a micro/premature optimization, but not worth spending more time nitpicking over :) LGTM!

brentpicasso: Yeah, it's a style / practice that I typically follow, unless it's only being referenced once or twice. "take the item off the shelf, use it, then put it back".

PR #15, opened 2015-02-19
V1.2.4 databus datafilters

brentpicasso:
connected bestlap and deltatime filters to view and fixed colors and stuff.

### DIFF
--- a/autosportlabs/racecapture/databus/databus.py
+++ b/autosportlabs/racecapture/databus/databus.py
@@ -18,6 +18,12 @@ class DataBusFactory(object):
 class DataBus(object):
     """Central hub for current sample data. Receives data from DataBusPump
     Also contains the periodic updater for listeners. Updates occur in the UI thread via Clock.schedule_interval
+    Architecture:    
+    (DATA SOURCE) => DataBus => (LISTENERS)
+    
+    Typical use:
+    (CHANNEL LISTENERS) => DataBus.addChannelListener()  -- listeners receive updates with a particular channel's value
+    (META LISTENERS) => DataBus.addMetaListener() -- Listeners receive updates with meta data
     """
     channel_metas = {}
     channel_data = {}
@@ -120,6 +126,10 @@ SAMPLE_POLL_EXCEPTION_RECOVERY = 10.0
 SAMPLES_TO_WAIT_FOR_META       = 5.0
 
 class DataBusPump(object):
+    """Responsible for dispatching raw JSON API messages into a format the DataBus can consume.
+    Attempts to detect asynchronous messaging mode, where messages are streamed to the DataBusPump.
+    If Async mode not detected, a polling thread is created to simulate this.
+    """
     _rc_api = None
     _data_bus = None
     sample = Sample()
--- a/autosportlabs/racecapture/databus/databus.py
+++ b/autosportlabs/racecapture/databus/databus.py
@@ -1,22 +1,31 @@
+from kivy.clock import Clock
 from time import sleep
 from threading import Thread, Event
 from autosportlabs.racecapture.data.sampledata import Sample, ChannelMeta, SampleMetaException,\
     ChannelMetaCollection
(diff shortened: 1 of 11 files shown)


autosportlabs/racecapture/databus/databus.py
```diff
@@ -1,22 +1,31 @@
...
+from autosportlabs.racecapture.databus.filter.laptimedeltafilter import LaptimeDeltaFilter
 
 DEFAULT_DATABUS_UPDATE_INTERVAL = 0.1 #10Hz UI update rate
 
+class DataBusFactory(object):
+    def create_standard_databus(self, system_channels):
+        databus = DataBus()
+        databus.add_data_filter(BestLapFilter(system_channels))
+        databus.add_data_filter(LaptimeDeltaFilter(system_channels))
+        return databus
```
ryandoherty: What's the benefit of having a DataBusFactory instead of other code just calling DataBus(system_channels) and the DataBus adding the data filters in its constructor?

ryandoherty:
80% of this is PEP8 changes, which look fine to me.

Took me a while to understand what functionality was added, I'm starting to think this codebase could benefit from a diagram or two. Or maybe more high-level comments on core classes that help define how data flows through the app (BT => DataBusPump? => DataBus => UI?).

ryandoherty:
Oh, and when creating a PR you put the description of what the changes do in the PR description, not a comment :)

brentpicasso:
Thanks, forgot to detail the PR as the train was pulling in. Yes, it's a
combination of PEP8 changes and some functionality.

On Wed, Feb 18, 2015 at 11:55 PM, Ryan Doherty <notifications@github.com>
wrote:

> Oh, and when creating a PR you put the description of what the changes do
> in the PR description, not a comment :)
>
> —
> <https://github.com/autosportlabs/RaceCapture_App/pull/15#issuecomment-75011152>
> .
>



-- 
Brent Picasso

autosportlabs/racecapture/databus/databus.py
brentpicasso: it's a little bit of decoupling as DataBus is not really aware of the filters it uses.  Other things add filters *to it*.

ryandoherty: I'm wondering about this because this is a factory pattern, but is so simple right now it doesn't appear to be a factory, just a convenience function to have default filters. 

Wondering if there's a better way. Maybe someone else with more CS experience can chime in to tell me if I'm barking up the wrong tree :)

brentpicasso: It fits the factory pattern fairly well; it's responsible for creating an instance of an object and configuring it in a particular way: http://en.wikipedia.org/wiki/Factory_%28object-oriented_programming%29

If you look at the DataBus class, you can see it's otherwise generic doesn't have anything racecapture/motorsports specific in it.  Admittedly, to fully realize this, it could live in a different namespace, like autosportlabs.databus instead of autosportlabs.racecapture.databus

ryandoherty: Yeah, I see your point about DataBus not having anything application/use case specific, which is good. I can now see why this was extracted (so the DataBus class is application agnostic). :+1:

ryandoherty:
:shipit: 

I'd still love a diagram or some of the critical classes documented structurally (not just one line sentences in a few functions) so the overall structure of the RC app can be grok'ed quickly. It's getting to be a reasonably complex app! (In a good way)

autosportlabs/racecapture/databus/databus.py
ryandoherty: I guess this discussion would be a good argument for having class-level comments that describe the `why

brentpicasso: Good feedback. Did some quick class level documentation describing the high level architecture of DataBus and DataBusPump.

PR #16, opened 2015-02-22
WIP: Fix scaling across all screen sizes

paulbutcher:
An in-progress pull request which will eventually (hopefully!) fix scaling across all screen sizes and densities

### DIFF
--- a/autosportlabs/racecapture/menu/homepageview.kv
+++ b/autosportlabs/racecapture/menu/homepageview.kv
@@ -16,14 +16,12 @@
 		BoxLayout:
 			orientation: 'vertical'
 			AnchorLayout:
-				size_hint_y: 0.1
-				size_hint_x: 1.0
+				size_hint: 1.0, 0.1
 				anchor_x: 'left'
 				anchor_y: 'center'
 
 				Image:
 					source: 'resource/images/app_icon_menu.png'
-					height: dp(85)
 					size_hint_x: None	
 			HLineSeparator:
 				size_hint_y: 0.05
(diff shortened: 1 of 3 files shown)


paulbutcher:
So far, I've only fixed the home screen, but I'm relatively pleased with the results. Here's what things looked like on a Nexus 4 (a small screen, high-density device) before:

nexus4 before

And here's what it looks like afterwards:

nexus4 after

And to show that we haven't broken things that used to work, here's what it used to look like on a 10in WXGA tablet before:

10in tablet before

And afterwards:

10in tablet after

Things may well require some tweaking, but I think it proves the basic approach. Let me know what you think.

autosportlabs/racecapture/menu/homepageview.kv
```diff
@@ -31,13 +28,13 @@
 				size_hint_y: 0.80
 				Label:
 					size_hint_y: None
-					font_size: dp(40)
+					font_size: self.height * 0.40
 					font_name: 'resource/fonts/ASL_regular.ttf'
```
brentpicasso: Let's use the pct_h() function in utils.py if possible, so we get the same concise effect as using dp() wherever we need to.

brentpicasso:
This looks really promising and what I hoped we'd get with the percent of height scaling.  I made one code comment which was to use the pct_h() function already in utils.py so we have a clean usage like dp(). 

Thanks!

brentpicasso:
still digging... to import functions in .kv files it requires a syntax like this:
#:import wh utils.wh

See "special syntaxes" section of http://kivy.org/docs/guide/lang.html

brentpicasso:
Dug further. I attempted to wrap these in helper functions and realized it defeated the fancy auto-property binding that kivy features. Even though it's more verbose, the scaling you came up with seems to be the best approach. 

we can still use the helper functions in utils.py to scale to window size when needed.

This is great, thank you Paul! it seems we can apply this approach to the dashboard views currently having problems, and then on to the config screens.

PR #18, opened 2015-03-02
Various trackmap fixes

brentpicasso:
* Added button to search for tracks; eliminated auto search with delay. also searches on enter/tab for desktop use
* Added first time run detection, prompt user to download tracks
* Added popup when tracks are loading after searching
* Various fixes to reflect better knowledge of kivy framework (replacing kvFind with self.ids.<widgetName>

### DIFF
--- a/autosportlabs/racecapture/api/rcpapi.py
+++ b/autosportlabs/racecapture/api/rcpapi.py
@@ -147,7 +147,7 @@ class RcpApi:
                 msg = comms.read_message()
                 if msg:
                     
-                    #print('msg_rx_worker Rx: ' + str(msg))
+                    print('msg_rx_worker Rx: ' + str(msg))
                     msgJson = json.loads(msg, strict = False)
                     self.on_rx(True)
                     error_count = 0
@@ -302,7 +302,7 @@ class RcpApi:
             comms = self.comms
 
             cmdStr = json.dumps(cmd, separators=(',', ':')) + '\r'
-            #print('send cmd: ' + cmdStr)
+            print('send cmd: ' + cmdStr)
             comms.write_message(cmdStr)
         except Exception:
             self.recover_connection()
--- a/autosportlabs/racecapture/api/rcpapi.py
+++ b/autosportlabs/racecapture/api/rcpapi.py
@@ -147,7 +147,7 @@ class RcpApi:
                 msg = comms.read_message()
                 if msg:
                     
-                    print('msg_rx_worker Rx: ' + str(msg))
+                    #print('msg_rx_worker Rx: ' + str(msg))
                     msgJson = json.loads(msg, strict = False)
                     self.on_rx(True)
                     error_count = 0
@@ -302,7 +302,7 @@ class RcpApi:
             comms = self.comms
 
             cmdStr = json.dumps(cmd, separators=(',', ':')) + '\r'
-            print('send cmd: ' + cmdStr)
+            #print('send cmd: ' + cmdStr)
             comms.write_message(cmdStr)
         except Exception:
             self.recover_connection()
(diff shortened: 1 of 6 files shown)


autosportlabs/racecapture/api/rcpapi.py
```diff
@@ -147,7 +147,7 @@ def msg_rx_worker(self):
                 msg = comms.read_message()
                 if msg:
                     
-                    #print('msg_rx_worker Rx: ' + str(msg))
+                    print('msg_rx_worker Rx: ' + str(msg))
```
ryandoherty: Purposefully leaving debug stuff here?

brentpicasso: well, fuck . :)

ryandoherty:
So I tested this, on startup I got the dialog box that asked to download track maps, now my app is stuck 'Searching'. I didn't search or do anything after clicking 'Ok'.

screenshot

I probably had some track maps already downloaded previously.

autosportlabs/racecapture/api/rcpapi.py
```diff
@@ -302,7 +302,7 @@ def sendCommand(self, cmd):
             comms = self.comms
 
             cmdStr = json.dumps(cmd, separators=(',', ':')) + '\r'
-            #print('send cmd: ' + cmdStr)
+            print('send cmd: ' + cmdStr)
             comms.write_message(cmdStr)
```
brentpicasso: both of these fixed. need to put this under logging.

ryandoherty:
I'm also still not to keen on popping up dialog boxes for things like this (especially on first run). It interrupts the user. Track maps could be downloaded in the background. Maybe we could have it run when the user clicks on the 'Race Tracks' button? (at least)

I'm worried that the more the app needs data from RCL/elsewhere the more dialog boxes we're going to throw around the app, making users frustrated. A 'firstrun' mode/check will definitely be useful

Technically the app can still function without track maps. IMHO modal dialogs are for when the app can't proceed and needs user input or must alert the user of specific information. 

Thoughts?

brentpicasso:
I've seen that too; it's an edge case if the app has detected first time
startup and there are no tracks to update.

Does it still happen if you re-start?

On Mon, Mar 2, 2015 at 7:44 AM, Ryan Doherty <notifications@github.com>
wrote:

> So I tested this, on startup I got the dialog box that asked to download
> track maps, now my app is stuck 'Searching'. I didn't search or do anything
> after clicking 'Ok'.
>
> [image: screenshot]
> <https://camo.githubusercontent.com/6f74e6f141216397d9d599c1a537ffd3f3fa2b37/687474703a2f2f692e696d6775722e636f6d2f773851357576712e706e67>
>
> I probably had some track maps already downloaded previously.
>
> —
> <https://github.com/autosportlabs/RaceCapture_App/pull/18#issuecomment-76734972>
> .
>



-- 
Brent Picasso

ryandoherty:
Also, after I restart the app I get the 'Searching' popup when I click 'Race Tracks'. It goes away, but since I didn't search for anything yet seems weird.

ryandoherty:
We should handle the case where the app has race tracks yet it detects a first time startup because we'll have customers upgrading from the old app to the new.

ryandoherty:
Also when I press enter/return the search box loses focus, making it hard to make corrections when searching.

brentpicasso:
I think on first startup it's ok to guide people through a number of steps, including this. Downloading  / updating stuff should happen in the background especially for browsing / loading race data from RCL.  We can circle back and do this for the track maps later if necessary.

brentpicasso:
re-focus search field if in desktop mode. Good stuff!

PR #19, opened 2015-03-06
Tire temp channels

brentpicasso:
*added channels for tire temperature with the following scheme
TireTmp_<corner><zone>
corner is LF,RF,LR,RR
zone is 1-4 where 1 = inner, 4 = outer

Also added BrakeTmp channels

Normalized Ride height channels to follow the scheme <Thing>_<location>

eg. Height_LF

### DIFF
--- a/resource/channel_meta/system_channels.json
+++ b/resource/channel_meta/system_channels.json
@@ -217,6 +217,42 @@
             "max": 2000
         },
         {
+            "nm": "BrakeTmp_LF",
+            "ut": "F",
+            "sys": 1,
+            "type": 1,
+            "prec": 0,
+            "min": 0,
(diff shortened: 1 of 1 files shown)


stieg:
Looks ok.  A side note is I don't like how we switch between imperial and metric.  Millimeters for the ride height but farenheight for the tire temps.

brentpicasso:
agreed. I think there's a point where we normalize on something - say metric system - and then go all in. Maybe on the next minor revision number.

ryandoherty:
LGTM!

stieg:
Not sure it would be a minor revision at that point.  Probably a major bump
since the full stack needs to be able to handle it.  But yes.

PR #21, opened 2015-03-14
Removed top level 'Race Tracks' menu items and changed cloud download icon

ryandoherty:
So there didn't seem to be any reason to have a high-level 'Race Tracks' menu item. It confused me and possibly other users as to how the top level tracks management related to the RCP track configuration. Since the Race Track/Sectors config handles everything on its own (including adding tracks to RCP) this seems simplest.

I also changed the cloud icon to a 'Download Tracks' button, improving usability. Tiny icons that don't look like buttons have no affordances, users won't know what it does or that it's clickable.

### DIFF
--- a/autosportlabs/racecapture/menu/homepageview.kv
+++ b/autosportlabs/racecapture/menu/homepageview.kv
@@ -41,16 +41,11 @@
 				icon: '\357\203\244'
 				title: 'Dashboard'
 				on_press: root.show_view('dash')
-			DisabledFeatureButton:
-				icon: '\357\202\200'
-				title: 'Analysis'
-				on_press: root.show_view('analysis')				
 			FeatureButton:
 				icon: '\357\202\205'
 				title: 'Configuration'
 				on_press: root.show_view('config')
-			FeatureButton:
-				icon: '\357\200\230'
-				title: 'Race Tracks'
-				on_press: root.show_view('tracks')
-			
+			DisabledFeatureButton:
+				icon: '\357\202\200'
+				title: 'Analysis'
+				on_press: root.show_view('analysis')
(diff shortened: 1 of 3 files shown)


brentpicasso:
I agree the top level race track browser is currently confusing. The original goal was to browse tracks, select one, and see interesting things about those tracks (who has the fastest lap, your friends lap, your laps and so on).  This could be part of the social component that lets you browse activity by people (friends) and by location (tracks). 

Might want to re-organize the home page icons to not leave a big empty space where that button was.

For the cloud Icon:

I really like the icon, and I understand they can be more discoverable. How about something like this wherever we use IconButtons?  Maybe make a flat button background similar to the Tile Icons we use on the home page?

image

ryandoherty:
I agree, it would be good to have icon+text. I did try to figure out how to do that, but I think because we're using fonts for icons it wasn't working. I think we'll have to copy how the mainmenu.kv file defines items and extend a button class for it.

brentpicasso:
Ok, let's work on adding text + icon. should we delete this PR or is there
something else in there you wanted to keep?

On Thu, Mar 19, 2015 at 11:57 AM, Ryan Doherty <notifications@github.com>
wrote:

> I agree, it would be good to have icon+text. I did try to figure out how
> to do that, but I think because we're using fonts for icons it wasn't
> working. I think we'll have to copy how the mainmenu.kv file defines items
> and extend a button class for it.
>
> —
> <https://github.com/autosportlabs/RaceCapture_App/pull/21#issuecomment-83718053>
> .
>



-- 
Brent Picasso

ryandoherty:
Yeah, we can close this PR for now.

PR #22, opened 2015-03-18
Changed all file system access to absolute paths to fix errors when bundling via Pyinstaller

ryandoherty:
tldr; we have to access all files absolutely, and it differs if we're in a bundled application vs running from source.

http://pythonhosted.org/PyInstaller/#adapting-to-being-frozen


brentpicasso:
Much good. Will get a pass after I double check it the windows installer.

ryandoherty:
:+1: This was a quick 'n dirty fix, so let me know if there's a better way to do this.

brentpicasso:
Tested this on windows. looks like it works well!

PR #24, opened 2015-03-25
V1.2.5 ui cleanup

brentpicasso:
Fixed a few things in the UI, plus properly implemented a dataFilter for handling displaying of lap times based on whether last lap time or predicted time is present in the stream.


ryandoherty:
Looks good to me. I don't have a way to test this, so I'll take your word for it that it's been tested enough :)

PR #26, opened 2015-03-26
simplified tachometer widget design into a sane, non-overengineered appr...

brentpicasso:
So. Tachometer widget was vastly over-designed to account for different RPM ranges. Simplified on one font and just honoring the channel metadata that comes down from RCP


ryandoherty:
I'll take any PR that removes more code than it adds ;)

PR #29, opened 2015-03-28
V1.3.0 trackmap

brentpicasso:
This is paired up with firmware PR https://github.com/autosportlabs/RaceCapture-Pro_firmware/pull/26

Trackmaps are now identified with a hard ID. This updates the behavior of the track manager and tracks viewer / selector as well as the API.

Trackmanager also will validate the on-disk trackmaps and if they aren't valid (missing createdAt field) it will not load from disk. An update will pull down new trackmaps.

### DIFF
--- a/autosportlabs/racecapture/config/rcpconfig.py
+++ b/autosportlabs/racecapture/config/rcpconfig.py
@@ -519,6 +519,7 @@ CONFIG_SECTOR_COUNT_STAGE = 18
 class Track(object):
     def __init__(self, **kwargs):
         self.stale = False
+        self.trackId = None
         self.trackType = TRACK_TYPE_CIRCUIT
         self.sectorCount = CONFIG_SECTOR_COUNT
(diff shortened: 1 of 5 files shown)


ryandoherty:
So, what happens if someone has updated their firmware but not the app or vice versa?

brentpicasso:
within same major version, a higher minor version firmware would be
backwards compatible with the app.

correspondingly, a newer version of the app won't work with an older minor
version of the firmware.

On Sat, Mar 28, 2015 at 9:54 PM, Ryan Doherty <notifications@github.com>
wrote:

> So, what happens if someone has updated their firmware but not the app or
> vice versa?
>
> —
> <https://github.com/autosportlabs/RaceCapture_App/pull/29#issuecomment-87358643>
> .
>



-- 
Brent Picasso

brentpicasso:
Closing to move on!

PR #30, opened 2015-03-29
added support for altitude and quality GPS channels

brentpicasso:
Added support for GPS channels altitude and GPS quality


ryandoherty:
LGTM!

PR #32, opened 2015-03-30
fixed parsing error where 32 and higher channels are present

brentpicasso:
App fails to parse data stream if >= 32 channels are present in the stream.

### DIFF
--- a/autosportlabs/racecapture/data/sampledata.py
+++ b/autosportlabs/racecapture/data/sampledata.py
@@ -119,7 +119,7 @@ class Sample(object):
             if mask_index >= 32:            
                 mask_index = 0;
                 bitmapIndex += 1
-                if bitmapIndex > bitmaskFields.length:
+                if bitmapIndex > len(bitmaskFields):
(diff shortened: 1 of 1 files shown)


brentpicasso:
You can test with this script, then watch the dashboard

setTickRate(10)
count = 20

A = {}
for v = 1, count do
A[v] = addChannel("Virtual" ..v, 10)
end

value = 0
function onTick()
 for i = 1, #A do
  setChannel(A[i], value * i)
 end
  value=value+1
  if value > 200 then value = 0 end
end

ryandoherty:
This is weird, I set the count to 33 and my dash looks like this:

![Dash]
(http://i.imgur.com/jtCQ4RT.png)

brentpicasso:
I put a trivial fix on this view to enable scrolling in master. I just merged it with this branch. Can you pull from this branch and test again?

ryandoherty:
Looks good!

![scroll]
(http://i.imgur.com/6t6byiq.png)

I do think there's a bug if there are too few channels the digital dash sizes things incorrectly thought.

PR #33, opened 2015-04-05
V2.8.0 channel meta synchronizations

brentpicasso:
re-request meta if we've re-connected- ensures metadata is in sync


brentpicasso:
auto-approving this PR and moving on b/c simple change.

PR #34, opened 2015-04-06
switched to a date utility package included with core python

brentpicasso:


utils.py
```diff
@@ -32,7 +35,10 @@ def boundary(value, minvalue, maxvalue):
     '''Limit a value between a minvalue and maxvalue'''
     return min(max(value, minvalue), maxvalue)
 
-
+def time_to_epoch(timestamp):
+    epoch = int(calendar.timegm(datetime.strptime(timestamp, "%Y-%m-%dT%H:%M:%SZ").timetuple()))
```
stieg: why even define epoch here?  Just make it a one liner or refactor the code a bit so its easier to read.

utils.py
```diff
@@ -9,12 +9,15 @@
            'is_color_transparent', 'boundary', 'dist',
            'deprecated', 'SafeList',
            'interpolate', 'OrderedDict', 'kvFind', 'kvFindClass', 'kvPrintAttr',
-           'breadth_first', 'walk_tree', 'filter_tree', 'kvquery', 'pct_h', 'pct_w')
+           'breadth_first', 'walk_tree', 'filter_tree', 'kvquery', 'pct_h', 'pct_w', 'time_to_epoch')
 
+import time
```
stieg: Do we organize imports alphabetically?  Also why some use of `from X import Y` syntax as opposed to being consistent?

autosportlabs/racecapture/tracks/trackmanager.py
```diff
@@ -4,7 +4,6 @@
 import errno
 import string
 import logging
-from dateutil.parser import *
```
stieg: Glad this is dying.

utils.py
brentpicasso: It was a habit from C / firmware code so it's easier to interactively debug. Don't do that much in python, it seems?

utils.py
brentpicasso: we don't ; it's semi arbitrarily organize it from core / high level packages to local imports, in that top to bottom order.

IDK on the other questions. seems some learning on importing python packages is in order.

utils.py
brentpicasso: Fixed / streamlined return.

PR #35, opened 2015-04-06
delta time should be in seconds

brentpicasso:


autosportlabs/racecapture/databus/filter/laptimedeltafilter.py
```diff
@@ -30,5 +31,5 @@ def filter(self, channel_data):
         reference_laptime = laptime if predtime == None else predtime
         delta_time = 0.0
         if reference_laptime and best_laptime:
-            delta_time = reference_laptime - best_laptime
+            delta_time = (reference_laptime - best_laptime) * 60.0
```
stieg: Wait... what is it in now?  If it were in Millis that this is wrong.  If it were already in seconds this would make it into minutes.

brentpicasso: The times are in decimal minutes and the channel units are in seconds.

Brent Picasso
Autosport Labs

stieg: Oh... duh.  Yeah that makes sense then :).

PR #38, opened 2015-04-27
V1.3.0 status api

ryandoherty:
This PR adds a status screen that shows the full status of the connected RCP. It polls RCP every 2 seconds to always have the latest status (due to things like GPS and cellular changing).

Todo: 
- ~~Figure out what icon to use in the side menu (and how to get the correct code to use it)~~
- I'd like to build a better 'Loading...' screen, but very minor nitpick
- Tests!


autosportlabs/racecapture/views/status/statusview.py
```diff
@@ -0,0 +1,288 @@
...
+
+    def __init__(self, track_manager, rc_api, **kwargs):
+        super(StatusView, self).__init__(**kwargs)
+        self.track_manager = track_manager
+        self.rc_api = rc_api
+        self.register_event_type('on_tracks_updated')
+        self.rc_api.addListener('status', self.on_status_updated)
+        self.menu_node = self.ids.menu
+        self.menu_node.bind(selected_node=self._on_menu_select)
+
```
brentpicasso: should prefix any of these with underscores if they're  meant as private only to class

autosportlabs/racecapture/views/status/statusview.py
```diff
@@ -0,0 +1,288 @@
...
+    selected_item = None
+
+    menu_built = False
+
+    #Track manager for getting track name
+    track_manager = None
+
+    #Used for building the left side menu
+    menu_keys = {
+        "system": "System Status",
```
brentpicasso: suggest _menu_keys here as well unless it's accessed outside of this class

autosportlabs/racecapture/views/status/statusview.py
```diff
@@ -0,0 +1,288 @@
...
+        "cell": "Cellular",
+        "bt": "Bluetooth",
+        "logging": "Logging",
+        "track": "Track",
+        "telemetry": "Telemetry"
+    }
+
+    #Dict for getting English text for status enums
+    enum_keys = {
+        'GPS': {
```
brentpicasso: ditto

autosportlabs/racecapture/views/status/statusview.py
ryandoherty: :+1:

autosportlabs/comms/comms.py
```diff
@@ -9,7 +9,7 @@
 COMMAND_KEEP_ALIVE = 'PING'
 
 def connection_process_message_reader(rx_queue, connection, should_run):
-    print('connection process message reader started')
+    #print('connection process message reader started')
```
brentpicasso: Was the commenting out of the logging info intentional or did it slip in a commit?

If you have time, might recommend adding in the Kivy Logging infrastructure as we've done here already: https://github.com/autosportlabs/RaceCapture_App/blob/master/service/main.py#L36

ryandoherty: I did comment it out during development b/c there's so much logging and it makes it hard to debug. I can add it back in if needed. 

I agree, I want to use a 'real' logging tool so we can handle it better.

brentpicasso: The kivy logging is super easy to implement, can almost be a search / replace.  only thing they want you to do is prefix it with the "area" being logged. e.g.  Logger.info('status_view:  or  Logger.info('comms:

ryandoherty: like

ryandoherty: Will split out logging stuff into a separate PR.

PR #40, opened 2015-05-04
V1.3.0 config view speedup

brentpicasso:
* Load .kv files later in the process
* Load screens only when they're selected
* optimize selection responsiveness in left tree

### DIFF
--- a/autosportlabs/racecapture/views/configuration/rcp/configview.py
+++ b/autosportlabs/racecapture/views/configuration/rcp/configview.py
@@ -130,7 +130,7 @@ class ConfigView(Screen):
         if not self.loaded:
             Builder.load_file(CONFIG_VIEW_KV)
             Builder.apply(self)
-            Clock.schedule_once(lambda dt: self.createConfigViews(),1.0)
+            self.createConfigViews()
         
     def createConfigViews(self):
         tree = kvFind(self, 'rcid', 'menu')
@@ -146,7 +146,8 @@ class ConfigView(Screen):
                 pass
     
             try:
-                if not value.view:
+                view = value.view
+                if not view:
                     view = value.view_builder()
                     self.configViews.append(view)
                     view.bind(on_config_modified=self.on_config_modified)
@@ -188,10 +189,9 @@ class ConfigView(Screen):
         attach_node('OBDII', None, lambda: OBD2ChannelsView(channels=runtime_channels, base_dir=self.base_dir))
         attach_node('Wireless', None, lambda: WirelessConfigView(base_dir=self.base_dir))
         attach_node('Telemetry', None, lambda: TelemetryConfigView())
-        attach_node('Scripting', None, create_scripting_view()
-)
+        attach_node('Scripting', None, lambda: create_scripting_view())
         if FIRMWARE_UPDATABLE:
-            attach_node('Firmware', None, FirmwareUpdateView(rc_api=self.rc_api, settings=self._settings))
+            attach_node('Firmware', None, lambda: FirmwareUpdateView(rc_api=self.rc_api, settings=self._settings))
         
         tree.bind(selected_node=on_select_node)
         tree.select_node(defaultNode)
--- a/autosportlabs/racecapture/views/configuration/rcp/configview.py
+++ b/autosportlabs/racecapture/views/configuration/rcp/configview.py
@@ -14,6 +14,7 @@ from kivy.uix.popup import Popup
 from kivy.uix.screenmanager import Screen
 from kivy.clock import Clock
 from kivy import platform
+from kivy.factory import Factory
 FIRMWARE_UPDATABLE =  not (platform == 'android' or platform == 'ios')
 
 from autosportlabs.racecapture.views.configuration.rcp.analogchannelsview import *
@@ -43,6 +44,7 @@ CONFIG_VIEW_KV = 'autosportlabs/racecapture/views/configuration/rcp/configview.k
 
 class LinkedTreeViewLabel(TreeViewLabel):
     view = None
+    view_builder = None
 
 class ConfigView(Screen):
     #file save/load
@@ -56,10 +58,9 @@ class ConfigView(Screen):
 
     #List of config views
     configViews = []
-    content = None
     menu = None
     rc_config = None
-    scriptView = None
+    script_view = None
     _settings = None
     base_dir = None
 
@@ -83,8 +84,6 @@ class ConfigView(Screen):
         self.register_event_type('on_poll_logfile')
         self.register_event_type('on_set_logfile_level')
         
-        self.content = kvFind(self, 'rcid', 'content')
-
     def on_config_written(self, *args):
         self.writeStale = False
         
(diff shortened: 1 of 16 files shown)


autosportlabs/racecapture/views/configuration/rcp/trackconfigview.py
```diff
@@ -18,7 +18,7 @@
 from autosportlabs.racecapture.views.configuration.baseconfigview import BaseConfigView
 from autosportlabs.racecapture.config.rcpconfig import *
 
-Builder.load_file('autosportlabs/racecapture/views/configuration/rcp/trackconfigview.kv')
+TRACK_CONFIF_VIEW_KV = 'autosportlabs/racecapture/views/configuration/rcp/trackconfigview.kv'
```
ryandoherty: Typo?

brentpicasso: fixed!

autosportlabs/racecapture/views/configuration/rcp/configview.py
```diff
@@ -137,50 +140,65 @@ def createConfigViews(self):
         def create_tree(text):
             return tree.add_node(LinkedTreeViewLabel(text=text, is_open=True, no_selection=True))
     
+        def show_node(node):
```
ryandoherty: I'm curious about this local function definition. Any particular reason it's done this way?

autosportlabs/racecapture/views/configuration/rcp/configview.py
```diff
@@ -137,50 +140,65 @@ def createConfigViews(self):
...
+                    if self.loaded:
+                        if self.config:
+                            view.dispatch('on_config_updated', self.config)
+                        if self.track_manager:
+                            view.dispatch('on_tracks_updated', self.track_manager)                                    
+                Clock.schedule_once(lambda dt: self.ids.content.add_widget(view))
+
+                
+            except Exception, e:
+                print e
```
ryandoherty: Maybe we should start using the Kivy Logger stuff from here on out? Also why a try/except? How could this throw an exception?

autosportlabs/racecapture/views/configuration/rcp/configview.py
```diff
@@ -137,50 +140,65 @@ def createConfigViews(self):
...
             except:
                 pass
-    
-            try:
-                self.content.clear_widgets()
-                Clock.schedule_once(lambda dt: self.content.add_widget(value.view))
-            except Exception, e:
-                print e
+            self.ids.content.clear_widgets()
+            Clock.schedule_once(lambda dt: show_node(value))
```
ryandoherty: So I understand this, why a schedule_once instead of just calling show_node?

autosportlabs/racecapture/views/configuration/rcp/configview.py
brentpicasso: IDK, the original example had this local function so I followed the pattern. Maybe it's a way to constrain scope?  Not sure if it's considered a best practice or not.

autosportlabs/racecapture/views/configuration/rcp/firmwareupdateview.py
```diff
@@ -19,7 +19,7 @@
 from time import sleep
 from threading import Thread
 
-Builder.load_file('autosportlabs/racecapture/views/configuration/rcp/firmwareupdateview.kv')
+FIRMWARE_UPDATE_VIEW_KV = 'autosportlabs/racecapture/views/configuration/rcp/firmwareupdateview.kv'
```
ryandoherty: Extra space at the end. (minor issue, but hey if we're here now...)

autosportlabs/racecapture/views/configuration/rcp/configview.py
brentpicasso: It schedules that call on the next frame, which let's the UI finish drawing for the current operation. Good to allow the user interface to be a bit more responsive. e.g. show that the left tree item was selected, then clear the window, then show the contents.

Without the schedule_once() the UI may lock up after the touch event before showing the item immediately selected.

autosportlabs/racecapture/views/configuration/rcp/gpiochannelsview.py
```diff
@@ -12,10 +12,11 @@
 from mappedspinner import MappedSpinner
 from kivy.metrics import dp
 
-Builder.load_file('autosportlabs/racecapture/views/configuration/rcp/gpiochannelsview.kv')
+GPIO_CHANNELS_VIEW_KV = 'autosportlabs/racecapture/views/configuration/rcp/gpiochannelsview.kv'
```
ryandoherty: Trailing whitespace.

autosportlabs/racecapture/views/configuration/rcp/obd2channelsview.py
```diff
@@ -10,7 +10,7 @@
 from utils import *
 from autosportlabs.racecapture.config.rcpconfig import *
 
-Builder.load_file('autosportlabs/racecapture/views/configuration/rcp/obd2channelsview.kv')
+OBD2_CHANNELS_VIEW_KV = 'autosportlabs/racecapture/views/configuration/rcp/obd2channelsview.kv'
```
ryandoherty: Trailing whitespace.

autosportlabs/racecapture/views/configuration/rcp/pwmchannelsview.py
```diff
@@ -13,10 +13,11 @@
 from kivy.metrics import dp
 from autosportlabs.racecapture.views.configuration.baseconfigview import BaseMultiChannelConfigView, BaseChannelView
 
-Builder.load_file('autosportlabs/racecapture/views/configuration/rcp/pwmchannelsview.kv')
+ANALOG_PULSE_CHANNELS_VIEW_KV = 'autosportlabs/racecapture/views/configuration/rcp/pwmchannelsview.kv'
```
ryandoherty: Trailing whitespace.

autosportlabs/racecapture/views/configuration/rcp/timerchannelsview.py
```diff
@@ -12,10 +12,11 @@
 from autosportlabs.racecapture.views.configuration.baseconfigview import BaseMultiChannelConfigView, BaseChannelView
 from autosportlabs.racecapture.config.rcpconfig import *
 
-Builder.load_file('autosportlabs/racecapture/views/configuration/rcp/timerchannelsview.kv')
+TIMER_CHANNELS_VIEW_KV = 'autosportlabs/racecapture/views/configuration/rcp/timerchannelsview.kv'
```
ryandoherty: Trailing whitespace.

autosportlabs/racecapture/views/configuration/rcp/wirelessconfigview.py
```diff
@@ -11,7 +11,7 @@
 from utils import *
 from autosportlabs.racecapture.views.configuration.baseconfigview import BaseConfigView
 
-Builder.load_file('autosportlabs/racecapture/views/configuration/rcp/wirelessconfigview.kv')
+WIRELESS_CONFIG_VIEW_KV = 'autosportlabs/racecapture/views/configuration/rcp/wirelessconfigview.kv'
```
ryandoherty: Trailing whitespace.

autosportlabs/racecapture/views/configuration/rcp/configview.py
brentpicasso: what it should really have is the Kivy logger and a error message popup. The try/except prevents the app from crashing out completely if something goes sideways here.

autosportlabs/racecapture/views/configuration/rcp/configview.py
ryandoherty: PEP8 doesn't mention nested functions and Google's Python standards (which I assume were at least reviewed by Guido) say it's ok. I guess it's ok, just curious what the benefits are.

autosportlabs/racecapture/views/configuration/rcp/configview.py
ryandoherty: What exception are we expecting? Seems a little weird to just catch everything and print it. I get that it will prevent the app from crashing but what conditions would get us to this situation and can we fix them instead of a top level catch?

IOW are there lower level things we can do for individual components to make them more resilient instead of catching all exceptions. 

If we're going to pop up an error message it should be very specific to whatever the issue is too.

autosportlabs/racecapture/views/configuration/rcp/configview.py
ryandoherty: :+1:

autosportlabs/racecapture/views/configuration/rcp/configview.py
brentpicasso: We have to do our best to prevent errors but we also need to prevent the app from crashing and literally exiting. That will happen with an uncaught exception.  It's different than a web server from showing a 500 but Apach keeps running - it would be like Apache crashing also.

ryandoherty: So does that mean we should move this to a top-level function in main.py instead? (seriously)

brentpicasso: Move what exactly?

ryandoherty: The try/except :)

brentpicasso: I think it's good here with an error dialog so we can say something like "Error showing config view XXX"

ryandoherty: Ok, sgtm.

But I don't think that messaging is useful to a user. If a config view (or anything in our app) throws an exception it should say something about why it failed so the user either can do something about it or if they send the message to use we can possibly understand it. Generic 'something went wrong' errors are not useful.

brentpicasso: Agreed. maybe we also can have a button that sends us a stack trace or copied everything to the clipboard so they can email  it to us

ryandoherty: :+1:

ryandoherty:
Looks good and the config section is way faster :) 

I would like to see the `print e` at least turned into a Kivy logger call.

brentpicasso:
Let's do that.

On Mon, May 4, 2015 at 11:32 AM, Ryan Doherty <notifications@github.com>
wrote:

> Looks good and the config section is way faster :)
>
> I would like to see the print e at least turned into a Kivy logger call.
>
> —
> <https://github.com/autosportlabs/RaceCapture_App/pull/40#issuecomment-98806130>
> .
>



-- 
Brent Picasso

brentpicasso:
removed print statements and added Kivy logger and alert dialog box.

ryandoherty:
Whitespace issues fixed?

brentpicasso:
removed trailing whitespace.

PR #43, opened 2015-05-07
minimum version check

brentpicasso:
Check for minimum firmware version and alerts user if detected version is incompatible. 

User will be able to update firmware and app will re-detect and re-connect.


ryandoherty:
Seems sane to me :)

PR #45, opened 2015-05-16
doing the lambda thing correctly

brentpicasso:
Correctly use lambda to schedule the status update on the next UI tick.

### DIFF
--- a/autosportlabs/racecapture/views/status/statusview.py
+++ b/autosportlabs/racecapture/views/status/statusview.py
@@ -128,7 +128,7 @@ class StatusView(Screen):
         self.track_manager = track_manager
         self.rc_api = rc_api
         self.register_event_type('on_tracks_updated')
-        self.rc_api.addListener('status', Clock.schedule_once(lambda dt: self._on_status_updated))
+        self.rc_api.addListener('status', self._on_status_updated)
         self._menu_node = self.ids.menu
         self._menu_node.bind(selected_node=self._on_menu_select)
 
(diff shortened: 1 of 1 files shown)


autosportlabs/racecapture/views/status/statusview.py
```diff
@@ -163,7 +163,10 @@ def _on_menu_select(self, instance, value):
         self.update()
 
     def _on_status_updated(self, status):
-        self.status = status['status']
+        def safe_update_status(status):
+            self.status = status['status']
```
ryandoherty: Is this how we've normally done this, a nested function? I swear I've seen us use addListener and pass a lambda. I'm mostly asking so I know how to do this properly in the future.

brentpicasso:
I just looked how the databus works and it's a completely disconnected process. with the Clock.schedule(). 

in main.py we have :

  def on_read_config_complete(self, rcpCfg):
        Clock.schedule_once(lambda dt: self.configView.dispatch('on_config_updated', self.rc_config))
        self.rc_config.stale = False
        self.showActivity('')

looking more..

ryandoherty:
For the config example, I think the status stuff is handled differently because the status attribute is a Kivy Property. That may be why things are different.

brentpicasso:
trying to see if it can be put in the .addListener call as was originally tried.

brentpicasso:
the current way is the cleanest I can do at the moment.

ryandoherty:
:shipit:

PR #46, opened 2015-05-16
V1.3.2 lapstats simplification

brentpicasso:
First cut at simplifying lap stats configuration view (now "Racing Stats") and logic for locking GPS sample rate to stats sample rate. 

Predicted time is handled separately. There's logic to force enabling of racing stats if pred time is enabled and to disable predicted time if lap stats are disabled.

### DIFF
--- a/autosportlabs/racecapture/config/rcpconfig.py
+++ b/autosportlabs/racecapture/config/rcpconfig.py
@@ -260,6 +260,7 @@ class LapConfigChannel(BaseChannel):
         return json_dict                
         
 class LapConfig(object):
+    DEFAULT_PREDICTED_TIME_SAMPLE_RATE = 5
     def __init__(self, **kwargs):
         self.stale = False
         self.lapCount = LapConfigChannel()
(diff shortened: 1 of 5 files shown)


autosportlabs/racecapture/config/rcpconfig.py
```diff
@@ -267,7 +268,29 @@ def __init__(self, **kwargs):
...
+        
+    def set_primary_stats(self, rate):
+        self.lapCount.sampleRate = rate
+        self.lapTime.sampleRate = rate
+        self.sector.sampleRate = rate
+        self.sectorTime.sampleRate = rate
+        self.elapsedTime.sampleRate = rate
+        self.currentLap.sampleRate = rate
+        self.stale = True
+
```
ryandoherty: I'm wondering if we should just put all the 'Racing Statistics' channels in an array to make our code simpler?

ryandoherty:
Overall looks good to me. I do wonder about the name 'Racing Statistics'. Naming things is hard. Other options:

- Lap Stats
- Timing
- Laptimes
- Lap/Run Timing

autosportlabs/racecapture/config/rcpconfig.py
brentpicasso: I think it is fine as it is. Even though we don't have static typing here
in python I think it's better to have them as properties on the object. And
I don't really see us adding to this soon.

Brent Picasso
Autosport Labs

PR #47, opened 2015-05-18
Make left side of menu icon actually work when clicked/tapped

ryandoherty:
The menu that slides out has a small overlap with the the menu icon, making it appear to not respond when tapped.


brentpicasso:
:shipit:

PR #48, opened 2015-05-18
Send empty track config if no tracks are selected in app

ryandoherty:
This fixes the bug where if you remove all tracks in the config screen in app and save the config to RCP, no tracks are actually removed.


brentpicasso:
looks good. make sure you update the changelog too.

PR #49, opened 2015-05-20
Use slightly different theme graphic so checkboxes show up on OSX.

ryandoherty:
This is the simplest and least amount of work needed for checkboxes to show up. Kivy's theming options are non-existent or absolute hacks that don't work well.


brentpicasso:
:shipit:

PR #53, opened 2015-05-25
V1.3.3 new bt connection

brentpicasso:
Removed background service, added BT connection as local java bridge


autosportlabs/comms/androidcomms.py
```diff
@@ -29,120 +15,48 @@ class PortNotOpenException(Exception):
...
     
     def isOpen(self):
-        return self._is_open
+        return self._bt_conn.isOpen()
     
     def open(self):
         print('Opening connection ' + str(self.port))
-        self.start_connection_process()
+        self._bt_conn.open(self.port)
+        print('after open!!!!!!!!!!!!!!!!')
```
ryandoherty: Only one exclamation point max per PEP8 standards.

brentpicasso: I was so excited

autosportlabs/racecapture/api/rcpapi.py
```diff
@@ -23,6 +23,7 @@
 
 COMMS_KEEP_ALIVE_TIMEOUT        = 2
 
+NO_DATA_AVAILABLE_DELAY         = 0.1
```
ryandoherty: Could we get a comment above this describing what it is? I can't tell from the name.

java/src/com/autosportlabs/racecapture/BluetoothConnection.java
```diff
@@ -0,0 +1,165 @@
...
+import java.util.UUID;
+import java.util.Set;
+import java.io.BufferedReader;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.OutputStream;
+import java.lang.reflect.Method;
+
+
+public class BluetoothConnection {
```
ryandoherty: Can we get a short description of why this class exists and what it does? An entire class/file with no comments makes me worry about future devs opening this. Also some comments for any particularly hairy or unique sections would help.

ryandoherty:
\o/ less code!

PR #57, opened 2015-06-01
correctly save files on android

brentpicasso:
* select the correct user data directory for android
* remove extraneous 'favorites' shortcuts in side bar

Future versions may simplify or create a different dialog for mobile.


autosportlabs/widgets/filebrowser/__init__.py
```diff
@@ -247,21 +248,22 @@ def fill_tree(self, fav_list):
...
-        places = ('Documents', 'Music', 'Pictures', 'Videos')
-        for place in places:
-            if isdir(user_path + place):
-                self.add_node(TreeLabel(text=place, path=user_path +
-                                        place), libs)
-        self._computer_node = self.add_node(TreeLabel(text='Computer',\
-        is_open=True, no_selection=True))
-        self._computer_node.bind(on_touch_down=self._drives_touch)
-        self.reload_drives()
+        if not is_mobile_platform():
```
ryandoherty: So what's the behavior if the platform is mobile?

brentpicasso: The android version just shows basically nothing in the left side list of 'favorites' - and also pre-selects the racecapture directory under external storage.

In the future we may want a simplified, mobile optimized file browser

PR #61, opened 2015-06-23
V1.3.5 auto detect hang

brentpicasso:
* make it so only the auto-detect worker can send messages to the API when in auto-detect mode. Prevents other parts of the system from crapping on the auto-detect worker while it's trying to do it's job

* Improve performance of auto-detection- thread synchronizations and preemptive checking of port status prevent multiple detection events to occur even if we've just finished detecting (avoid race conditions)

* Simplified auto-detection message sending and receiving

* Prevent the UI thread from hanging forever in case the command TX queue is full

* improve performance of auto detection in Linux by ignoring ttyUSB* ports

* Cleaned up logging

### DIFF
--- a/autosportlabs/comms/comms.py
+++ b/autosportlabs/comms/comms.py
@@ -86,6 +86,7 @@ def connection_message_process(connection, port, rx_queue, tx_queue, command_que
 class Comms():
     CONNECT_TIMEOUT = 1.0    
     DEFAULT_TIMEOUT = 1.0
+    QUEUE_FULL_TIMEOUT = 1.0
     _timeout = DEFAULT_TIMEOUT
     port = None
     _connection = None
@@ -151,5 +152,5 @@ class Comms():
     
     def write_message(self, message):
         if not self.isOpen(): raise Exception('Comms Exception')
-        self._tx_queue.put(message, True, 1.0 )
+        self._tx_queue.put(message, True, Comms.QUEUE_FULL_TIMEOUT )
                     
--- a/autosportlabs/comms/comms.py
+++ b/autosportlabs/comms/comms.py
@@ -10,7 +10,7 @@ COMMAND_CLOSE      = 'CLOSE'
 COMMAND_KEEP_ALIVE = 'PING'
 
 def connection_process_message_reader(rx_queue, connection, should_run):
-    Logger.info('Comms: connection process message reader started')
+    print('connection process message reader started')
     
     while should_run.is_set():
         try:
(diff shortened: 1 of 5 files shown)


autosportlabs/comms/comms.py
```diff
@@ -3,13 +3,14 @@
 import multiprocessing
 from Queue import Empty
 from time import sleep
+from kivy.logger import Logger
```
blalor: Where does this output go in production builds?

autosportlabs/comms/comms.py
```diff
@@ -131,24 +132,24 @@ def cleanup(self):
...
         try:
             return self._rx_queue.get(True, self._timeout)
         except: #returns Empty object if timeout is hit
             return None
     
     def write_message(self, message):
-        if not self.isOpen(): raise Exception("Comms Exception")
-        self._tx_queue.put(message)
+        if not self.isOpen(): raise Exception('Comms Exception')
+        self._tx_queue.put(message, True, 1.0 )
```
brentpicasso: this prevents the UI thread from possibly being blocked  if the Queue is full.

autosportlabs/comms/comms.py
brentpicasso: It's a kivy thing. I don't actually know :-|  I think there's a way to capture this to a file. Need to research it. But, first step is to log stuff the right way

autosportlabs/comms/comms.py
stieg: May want to comment on the meaning of the extra params.

autosportlabs/racecapture/api/rcpapi.py
```diff
@@ -174,10 +177,10 @@ def msg_rx_worker(self):
                 sleep(1.0)
             except Exception:
                 Logger.warn('RCPAPI: Message rx worker exception: {} | {}'.format(msg, str(Exception)))
-                traceback.print_exc()
+                Logger.debug(traceback.format_exc())
                 msg = ''
                 error_count += 1
-                if error_count > 5:
+                if error_count > 5 and not self._auto_detect_event.is_set():
```
brentpicasso: this prevents the message rx worker from triggering a re-auto detect if we're in the middle of... auto-detecting.

autosportlabs/racecapture/api/rcpapi.py
```diff
@@ -614,21 +617,20 @@ class VersionResult(object):
...
-        
+                    
         while True:
             try:
                 self._auto_detect_event.wait()
                 self._auto_detect_event.clear()
                 self._enable_autodetect.wait()
                 Logger.info("RCPAPI: Starting auto-detect")
+                self._auto_detect_busy.set()
+                self.sendCommandLock.acquire()
```
brentpicasso: this grabs the recursive lock to prevent other parts of the app from chatting with the API

autosportlabs/comms/comms.py
brentpicasso: I'll remove the magic number (1.0), which should help.

autosportlabs/racecapture/api/rcpapi.py
ryandoherty: Comments explaining the why of things would even better if they were in the source code for future developers to see. Especially if it's something tricky like this.

brentpicasso: :+1:

PR #64, opened 2015-06-30
V1.3.6 loading performance improvements

brentpicasso:
Results in about 2-3 seconds faster load time on my newer phone.

* lazy load main screens to speed up load time
* ESC / back button doesn't do a "double back" if dismissing a dialog
* some clean up of main.py
* factory functions for creating main screens
* do fewer things in build() to speed up load time
* decouple status_view polling into it's own status_pump() object, so status data can be used elsewhere (top status bar, for example)
* make various views work correctly in the context of lazy loading
* PEP8 corrections
* converted to use of Logger in key areas

### DIFF
--- a/autosportlabs/racecapture/status/statuspump.py
+++ b/autosportlabs/racecapture/status/statuspump.py
@@ -1,0 +1,46 @@
+import kivy
+kivy.require('1.8.0')
+from kivy.logger import Logger
+from kivy.clock import Clock
+from time import sleep
+from threading import Thread
+
+
+
+"""Responsible for querying status from the RaceCapture API
+"""
+class StatusPump(object):
+    
+    #how often we query for status
+    STATUS_QUERY_INTERVAL = 2.0
+    
+    #Connection to the RC API
+    _rc_api = None
+    
+    #Things that care about status updates
+    _listeners = []
+    
+    #Worker Thread
+    _status_thread = None
+    
+    def add_listener(self, listener):
+        self._listeners.append(listener)
+                    
+    def start(self, rc_api):
+        Logger.info('StatusPump: starting')
+        self._rc_api = rc_api
+        self._status_thread = Thread(target=self.status_worker)
+        self._status_thread.daemon = True
+        self._status_thread.start()
+        
+    def status_worker(self):
+        self._rc_api.addListener('status', self._on_status_updated)        
+        while True:
+            self._rc_api.get_status()
+            sleep(self.STATUS_QUERY_INTERVAL)
+        
+    def _on_status_updated(self, status):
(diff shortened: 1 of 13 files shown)


brentpicasso:
this branch should really be called 1.3.6_loading_performance_improvements...

autosportlabs/racecapture/status/statuspump.py
```diff
@@ -0,0 +1,46 @@
...
+    #Things that care about status updates
+    _listeners = []
+    
+    #Worker Thread
+    _status_thread = None
+    
+    def add_listener(self, listener):
+        self._listeners.append(listener)
+                    
+    def start(self, rc_api):
```
ryandoherty: Do we need a way to stop querying status? What happens if RCP is not connected?

autosportlabs/racecapture/status/statuspump.py
```diff
@@ -0,0 +1,46 @@
...
+    #Connection to the RC API
+    _rc_api = None
+    
+    #Things that care about status updates
+    _listeners = []
+    
+    #Worker Thread
+    _status_thread = None
+    
+    def add_listener(self, listener):
```
ryandoherty: We may also need a way to remove a listener, especially if we start destroying views or widgets for performance reasons.

main.py
```diff
@@ -257,22 +258,96 @@ def _getX(self):
...
+    def build_preferences_view(self):
+        preferences_view = PreferencesView(name='preferences', settings=self.settings, base_dir=self.base_dir)
+        return preferences_view
+    
+    def build_homepage_view(self):
+        homepage_view = HomePageView(name='home')
+        homepage_view.bind(on_select_view = lambda instance, view_name: self.switchMainView(view_name))
+        return homepage_view
+
+    def init_viewbuilders(self):
```
ryandoherty: Pedantic: seems like this should be named `init_view_builders`

ryandoherty:
Looks good! Love the use of Kivy's logger.

autosportlabs/racecapture/status/statuspump.py
brentpicasso: Since it's a daemon thread the thread automatically exits when the app exits. 

Didn't have a use case for stopping querying while app is running.

There's no harm with commands being sent with RC disconnected; they fail gracefully within the comms layer.

autosportlabs/racecapture/status/statuspump.py
brentpicasso: I thought about this too, will add a remove_listener when we first need it.

main.py
brentpicasso: :+1:

PR #71, opened 2015-07-09
New build and packaging tools for OSX

ryandoherty:
Pyinstaller sucks, new Kivy 1.9.0 sucks at other things. This is the 'new', 'recommended' build process for OSX apps.

Most of this is just copying everything in the Kivy runtime and our app into a folder, renaming it, adding some OSX goodies so it 'looks' like a real app and packaging it up into a DMG file.


brentpicasso:
:shipit:

PR #73, opened 2015-08-02
V1.3.7 script view windows fix

brentpicasso:
ripple effect from 1.9.0 conversion.  pyinstaller/pygments hack is still needed for windows build, but now we can just use the pygments supplied as a standard library dependency, as it should be.


ryandoherty:
:shipit:

PR #76, opened 2015-08-22
App telemetry

ryandoherty:
100% complete on sending app telemetry to RCL. 

Architecture is TelemetryManager create TelemetryConnection in a new thread. Events bubble up from connection to manager to app. 

App will continually reconnect every 2s when connections fail.


autosportlabs/racecapture/databus/databus.py
```diff
@@ -29,22 +29,30 @@ class DataBus(object):
...
     channel_metas = {}
     channel_data = {}
+    sample = None
     channel_listeners = {}
     meta_listeners = []
     meta_updated = False
     data_filters = []
-    
+    sample_listeners = []
+    _polling = False
```
stieg: NIT: Inconsistent naming.  Any reason?

autosportlabs/racecapture/databus/databus.py
```diff
@@ -29,22 +29,30 @@ class DataBus(object):
...
+
     def __init__(self, **kwargs):
         super(DataBus, self).__init__(**kwargs)
 
     def start_update(self, interval = DEFAULT_DATABUS_UPDATE_INTERVAL):
-        Clock.schedule_interval(self.notify_listeners, interval)
+        if not self._polling:
+            Clock.schedule_interval(self.notify_listeners, interval)
+            self._polling = True
```
stieg: Seems redundant no?  If we got in this condition then self._polling was already True

stieg: Nevermind... I misread this.  Keep it :)

stieg: Consider this:
```python
if self._polling:
    return
Clock.schedule_interval(self.notify_listeners, interval)
self._polling = True
```
The negative case like this saves indents and makes flow easier.

autosportlabs/racecapture/databus/databus.py
```diff
@@ -29,22 +29,30 @@ class DataBus(object):
...
 
     def stop_update(self):
         Clock.unschedule(self.notify_listeners)
```
stieg: What happens if this is called multiple times, even if the clock is stopped?

autosportlabs/telemetry/telemetryconnection.py
```diff
@@ -0,0 +1,396 @@
...
+from kivy.event import EventDispatcher
+import threading
+import asynchat, asyncore
+import json
+import socket
+import sys
+import errno
+import math
+
+#  Manager that creates a new telemetry connection in a separate thread
```
stieg: I thought file description headers were to go into blocks like this:
```python
"""
HI

I describe this file
"""
```

autosportlabs/telemetry/telemetryconnection.py
```diff
@@ -0,0 +1,396 @@
...
+
+        self._data_bus.addMetaListener(self._on_meta)
+        self._data_bus.start_update()
+
+        meta = self._data_bus.getMeta()
+
+        if len(meta) > 1:
+            self.channels = meta
+        else:
+            self.channels = None
```
stieg: Could use a python ternary statement here:
```python
self.channels = meta if len(meta) > 1 else None
```
Then again... do you even need that as the default value of self.channels (if never initialized) should already be None right?

stieg: That last sentence is wrong.  Ignore it.

autosportlabs/telemetry/telemetryconnection.py
```diff
@@ -0,0 +1,396 @@
...
+
+        if host is not None:
```
stieg: Huh?  This is useless based on what happening in line 27 no?

autosportlabs/telemetry/telemetryconnection.py
```diff
@@ -0,0 +1,396 @@
...
+            self.host = host
+
+        if port is not None:
```
stieg: Ditto

autosportlabs/telemetry/telemetryconnection.py
```diff
@@ -0,0 +1,396 @@
...
+
+        if 'auto_start' in kwargs:
+            self.auto_start = kwargs.get('auto_start')
+
+        if self.auto_start:
+            self.start()
+
+    # Event handler for when meta (channel list) changes
+    def _on_meta(self, channel_metas):
+        Logger.info("TelemetryManager: Got meta")
```
stieg: Should this be `debug` instead of `info`?

autosportlabs/telemetry/telemetryconnection.py
```diff
@@ -0,0 +1,396 @@
...
+    # Event handler for when config is written to RCP
+    def on_config_written(self, config):
+        self.device_id = config.connectivityConfig.telemetryConfig.deviceId
+
+    # Starts connection, checks to see if requirements are met
+    def start(self):
+        Logger.info("TelemetryManager: start()")
+        if self._connection_process:
+            # We have a connection object, could be alive or dead
+            if self._connection_process.is_alive():
```
stieg: Simplify
```python
if not self._connection_process.is_alive():
    self._connect()
```

autosportlabs/telemetry/telemetryconnection.py
```diff
@@ -0,0 +1,396 @@
...
+class TelemetryConnection(asynchat.async_chat):
+
+    STATUS_UNINITIALIZED = -1
+    STATUS_DISCONNECTED = 0
+    STATUS_CONNECTED = 1
+    STATUS_AUTHORIZED = 2
+    STATUS_STREAMING = 3
+
+    ERROR_CONNECTING = 0
+    ERROR_AUTHENTICATING = 1
```
stieg: NIT: standard protocol is that UNKNOWN error is always the first error code.  Just a note more than anything :)

autosportlabs/telemetry/telemetryconnection.py
```diff
@@ -0,0 +1,396 @@
...
+        elif status_code == TelemetryConnection.STATUS_STREAMING:
+            self.dispatch('on_streaming', msg)
+        elif status_code in [TelemetryConnection.ERROR_AUTHENTICATING,
+                             TelemetryConnection.ERROR_CONNECTING,
+                             TelemetryConnection.ERROR_UNKNOWN,
+                             TelemetryConnection.ERROR_UNKNOWN_MESSAGE]:
+            self.dispatch('on_error', msg)
+
+    def on_connected(self, *args):
+        pass
```
stieg: Log messages that you got here?

autosportlabs/telemetry/telemetryconnection.py
```diff
@@ -0,0 +1,396 @@
...
+            self._send_meta()
+            self._start_sample_timer()
+
+    # Sets up timer to send data to RCL every 100ms
+    def _start_sample_timer(self):
+        self._sample_timer = threading.Timer(0.1, self._send_sample)
+        self._sample_timer.start()
+
+    def run(self):
+        Logger.info("TelemetryConnection: connecting to: " + self.host + " " + str(self.port))
```
stieg:
```python
Logger.info("TelemetryConnection: connecting to %s:%d" % (self.host, self.port))
```

autosportlabs/telemetry/telemetryconnection.py
```diff
@@ -0,0 +1,396 @@
...
+
+        # No try/except here because the connect call ends up calling socket.connect_ex,
+        # which does not throw any errors. Async FTW! (Sorta)
+        self.create_socket(socket.AF_INET, socket.SOCK_STREAM)
+        self.connect((self.host, self.port))
+
+        asyncore.loop()
+
+    def close_connection(self):
+        self.close_when_done()
```
stieg: Logging?

autosportlabs/telemetry/telemetryconnection.py
```diff
@@ -0,0 +1,396 @@
...
+        if hasattr(v, 'errno'):
+            if v.errno == errno.ECONNREFUSED:
+                Logger.info("TelemetryConnection: connection refused")
+                self._update_status("error", "Connection refused", self.ERROR_CONNECTION_REFUSED)
+            elif v.errno == errno.ETIMEDOUT:
+                Logger.info("TelemetryConnection: timeout connecting")
+                self._update_status("error", "Timeout connecting", self.ERROR_TIMEOUT)
+            else:
+                Logger.info("TelemetryConnection: unknown error connecting " + str(t) + " " +str(v))
+                self._update_status("error", "Timeout connecting", self.ERROR_UNKNOWN)
```
stieg: Copy pasta error?

autosportlabs/telemetry/telemetryconnection.py
```diff
@@ -0,0 +1,396 @@
...
+        else:
```
stieg: Seems like this can be merged with the above if/else block with a bit of careful handling of t and v variables.

runtests.py
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/python
 import unittest
+#from kivy.logger import Logger
```
stieg: Cleanup?

autosportlabs/racecapture/databus/databus.py
ryandoherty: Underscores are to denote private variables/functions if that's what you are referring to.

autosportlabs/racecapture/databus/databus.py
ryandoherty: :+1:

autosportlabs/racecapture/databus/databus.py
ryandoherty: ¯\_(ツ)_/¯

autosportlabs/telemetry/telemetryconnection.py
ryandoherty: Yup

autosportlabs/telemetry/telemetryconnection.py
ryandoherty: ¯\\\_(ツ)_/¯

autosportlabs/telemetry/telemetryconnection.py
ryandoherty: :+1:

autosportlabs/telemetry/telemetryconnection.py
ryandoherty: :+1:

autosportlabs/telemetry/telemetryconnection.py
ryandoherty: Thoughts on ternary operators for all the keyword args that are then copied to instance vars? I'm not a big fan of Python's ternary operators, which is why I used the if blocks instead.

main.py
```diff
@@ -426,5 +434,43 @@ def rc_detect_activity(self, info):
...
+        host = self.getAppArg('telemetryhost')
+
+        self._telemetry_connection = TelemetryManager(self._databus, host=host)
+        self.config_listeners.append(self._telemetry_connection)
+        self._telemetry_connection.bind(on_connected=self.telemetry_connected)
+        self._telemetry_connection.bind(on_disconnected=self.telemetry_disconnected)
+        self._telemetry_connection.bind(on_streaming=self.telemetry_streaming)
+        self._telemetry_connection.bind(on_error=self.telemetry_error)
+
+        if self.settings.userPrefs.get_pref('preferences', 'send_telemetry') == "1":
```
brentpicasso: is send_telemetry like a global disable switch that can disable it regardless of the RCP setting?

toolbarview.kv
```diff
@@ -50,18 +50,12 @@
 	    	size_hint_x: 0.5
 			rcid: 'pbar'
 	    IconButton:
-	    	rcid: 'teleTxStatus'
+	    	rcid: 'teleStatus'
```
brentpicasso: so just a single activity LED? I think that'll be OK, even though it'll be nice to see activity of messages coming from RCL

autosportlabs/telemetry/telemetryconnection.py
brentpicasso: +1 on debug

autosportlabs/telemetry/telemetryconnection.py
```diff
@@ -0,0 +1,420 @@
...
+
+    # Event handler for when meta (channel list) changes
+    def _on_meta(self, channel_metas):
+        Logger.info("TelemetryManager: Got meta")
+        self.channels = channel_metas
+
+    # Event handler for when self.channels changes, don't restart connection b/c
+    # the TelemetryConnection object will handle new channels
+    def on_channels(self, instance, value):
+        Logger.info("TelemetryManager: Got channels")
```
brentpicasso: suggest debug

autosportlabs/telemetry/telemetryconnection.py
```diff
@@ -0,0 +1,420 @@
...
+        if not self._connection_process and self.auto_start:
+            self.start()
+
+    # Event handler for when self.device_id changes, need to restart connection
+    def on_device_id(self, instance, value):
+        # Disconnect, re-auth, etc
+        Logger.info("TelemetryManager: Got new device id")
```
brentpicasso: ditto

autosportlabs/telemetry/telemetryconnection.py
```diff
@@ -0,0 +1,420 @@
...
+    def on_config_updated(self, config):
+        self.device_id = config.connectivityConfig.telemetryConfig.deviceId
+
+    # Event handler for when config is written to RCP
+    def on_config_written(self, config):
+        self.device_id = config.connectivityConfig.telemetryConfig.deviceId
+
+    # Starts connection, checks to see if requirements are met
+    def start(self):
+        Logger.info("TelemetryManager: start()")
```
brentpicasso: most of these could be debug, except for book-ending major operations.

autosportlabs/telemetry/telemetryconnection.py
```diff
@@ -0,0 +1,420 @@
...
+
+            if self._sample_timer:
+                self._sample_timer.cancel()
+
+            self._send_meta()
+            self._start_sample_timer()
+
+    # Sets up timer to send data to RCL every 100ms
+    def _start_sample_timer(self):
+        self._sample_timer = threading.Timer(0.1, self._send_sample)
```
brentpicasso: extract magic number into constant or preferences value

autosportlabs/telemetry/telemetryconnection.py
```diff
@@ -0,0 +1,420 @@
...
+            self._update_status("ok", "Connected to RaceCapture/Live", self.STATUS_CONNECTED)
+            self._connected = True
+            self._connecting = False
+            self._send_auth()
+
+    def handle_expt(self):
+        # Something really bad happened if we're here
+        Logger.info("TelemetryConnection: handle_expt, closing connection")
+        self._update_status("ok", "Unknown error, disconnected from RaceCapture/Live", self.STATUS_DISCONNECTED)
+        self.close_when_done()
```
brentpicasso: If we have repeated connect failures, does it retry aggressively or does it back off / sleep for a bit? Concern is flooding RCL and/or consume a lot of CPU in the app.

main.py
```diff
@@ -426,5 +434,43 @@ def rc_detect_activity(self, info):
...
+
+    def telemetry_error(self, instance, msg):
+        self.showActivity(msg)
+
+    def on_config_change(self, config, section, key, value):
+        """Called any time the app preferences are changed
+        """
+        token = (section, key)
+
+        if token == ('preferences', 'send_telemetry'):
```
brentpicasso: Ok, we're using the device ID from RC device, but using an in-app preference for controlling whether we send telemetry or not. Is there a way we can have RC config be the single source of truth for controlling whether telemetry is sent?  ..thinking of what switches and settings can be inferred. background streaming? Might require an API change on RC to indicate when things should be logging or not.

toolbarview.kv
ryandoherty: Yeah, only one as having 2 would confuse a user. A user would only want to know if a connection was made and it was working. Having an icon that blinks when messages are sent down from RCL  isn't really that useful, if we need to see them we'll fire up the app and view the debug output.

autosportlabs/telemetry/telemetryconnection.py
ryandoherty: :+1:

autosportlabs/telemetry/telemetryconnection.py
ryandoherty: :+1:

autosportlabs/telemetry/telemetryconnection.py
ryandoherty: See https://github.com/autosportlabs/RaceCapture_App/pull/76/files#diff-1270f672312ff41def87cdc098b7e1d0R135

App retries every 2 seconds. Hadn't thought about backing off, part of me wants the app to try to connect again quickly. I think an exponential backoff, but limiting the max wait to 10s would work. Thoughts?

main.py
ryandoherty: I thought about this a bit, and think this being an app preference is best for the short term for a few reasons.

1) No change to RC API
2) Same behavior as before
3) The streaming from the app is functionality that is *very* specific to the app, not to any other system RCP may talk to. 

If we attempt to make this work with the current RCP configuration, I think that will cause problems and confusion by attempting to 'infer' what the user may want to do via configuration settings that aren't related to streaming via the app. 

Long-term we need to update the telemetry part of the configuration for RCP to have a way to select how a user wants telemetry sent to RCL: built-in telemetry module or app.

brentpicasso: What about this scenario:

* Telemetry device ID is specified
* Cellular module is set to 'off'

This could be inferred as "I wan to stream telemetry".

We could still have an app preferences switch, enabled by default, so they'd have to go in to disable it, if they want to. 

Later we can introduce the logic for enabling telemetry with different rules, like we've talked about for logging - speed based, RPM based, etc.

ryandoherty: I'm confused. If we have an app preference for streaming, why attempt to infer based on a user turning off their cellular module? Many users turn off cellular to save $ on their SIM card, myself included.

Implicit, hidden behavior will result in a poor user experience. KISS

brentpicasso: The primary benefit from inferring the state is to prevent a scenario where  a user could have a cellular module, with SIM card and then enables telemetry in the app preferences, thinking that's what they should be doing.  We don't want a connected phone *and* RCP's cellular module sending telemetry at the same time.

If we infer the settings we can actually make for a simple experience for the user who has an RCP with no cellular module / or has it turned off.  All they would have to do is enter the device id and it would start working.

ryandoherty: The problem with that behavior is the app will broadcast data to RCL any time it connects to a RCP that has a device ID and cellular turned off. A user would not expect that behavior. Safe defaults are best. 

If we go that route when we push out this feature anyone with a RCP with a device ID and telemetry turned off will immediately start broadcasting live data even though the user doesn't know this is happening. In fact, if the user had turned off telemetry, it means they *don't* want live data broadcasted.

This is unexpected behavior that has big potential for privacy concerns, user outrage and confusion.

brentpicasso: This is fine. In the future we can step a user through a process where they choose how they want telemetry enabled.   So, the app-based telemetry default setting should be 'off' as you have it now.

However, IMO the app still would need to look at the cellular module setting so we won't create duplicate telemetry sessions- i.e. avoid them enabling app telemetry *and* enabling RCP telemetry simultaneously. I don't think this is an edge case and can cause various thorny issues, both technically and user experience.

ryandoherty: Good point about the double connection issue, will have the app not send telemetry if the cellular module is on.

brentpicasso:
Bump version to 1.3.8 and update CHANGELOG and I think we're close to merging after some more testing!

PR #77, opened 2015-08-27
Use sending thread for telemetry data

ryandoherty:


autosportlabs/telemetry/telemetryconnection.py
```diff
@@ -263,15 +265,23 @@ def _on_meta(self, meta):
...
 
     # Sets up timer to send data to RCL every 100ms
     def _start_sample_timer(self):
-        self._sample_timer = Clock.schedule_interval(self._send_sample, self.SAMPLE_INTERVAL)
+        self._running.set()
+        self._sample_timer = threading.Thread(target=self._sample_worker)
+        self._sample_timer.daemon = True
+        self._sample_timer.start()
+
+    def _sample_worker(self):
```
brentpicasso: I would wrap this with a try / except to see if something wayward happens. And add logging for entry / exit events

autosportlabs/telemetry/telemetryconnection.py
```diff
@@ -263,15 +265,23 @@ def _on_meta(self, meta):
...
+        while self._running.is_set():
+            self._send_sample()
```
brentpicasso: actually try / except can go here

autosportlabs/telemetry/telemetryconnection.py
```diff
@@ -263,15 +266,26 @@ def _on_meta(self, meta):
...
+        self._sample_timer.daemon = True
+        self._sample_timer.start()
+
+    def _sample_worker(self):
+        while self._running.is_set():
+            try:
+                self._send_sample()
+                sleep(self.SAMPLE_INTERVAL)
+            except Exception as e:
+                Logger.error("TelemetryConnection: error sending sample: " + str(e))
```
brentpicasso: so here we're not closing the connection (calling end() ) if we get an exception.  How would we handle an error when sending data? Would we be relying on a different mechanism?

brentpicasso:
:shipit:

PR #78, opened 2015-08-27
telemetry icon status colors; a few refinements + bug fix

brentpicasso:
added telemetry icon colors

guarded against none object for thread join

made toolbar message more concise

### DIFF
--- a/autosportlabs/telemetry/telemetryconnection.py
+++ b/autosportlabs/telemetry/telemetryconnection.py
@@ -38,6 +38,7 @@ class TelemetryManager(EventDispatcher):
         self._retry_count = 0
         self._auth_failed = False
 
+        self.register_event_type('on_connecting')
         self.register_event_type('on_connected')
         self.register_event_type('on_disconnected')
         self.register_event_type('on_streaming')
@@ -141,6 +142,7 @@ class TelemetryManager(EventDispatcher):
     # Creates new TelemetryConnection in separate thread
     def _connect(self):
         Logger.info("TelemetryManager: starting connection")
+        self.dispatch('on_connecting', "Connecting to RaceCapture/Live")
         self.connection = TelemetryConnection(self.host, self.port, self.device_id,
                                               self.channels, self._data_bus, self.status)
         self._connection_process = threading.Thread(target=self.connection.run)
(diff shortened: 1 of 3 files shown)


autosportlabs/telemetry/telemetryconnection.py
```diff
@@ -173,7 +175,7 @@ def status(self, status, msg, status_code):
             Logger.warning("TelemetryManager: authentication failed")
             self._auth_failed = True
             self.stop()
-            self.dispatch('on_auth_error', "Error connecting to RaceCapture/Live, invalid device id")
+            self.dispatch('on_auth_error', "RaceCapture/Live: invalid device id")
```
brentpicasso: Made this more concise; line wrapped on my phone form factor.

autosportlabs/telemetry/telemetryconnection.py
```diff
@@ -486,6 +491,7 @@ def end(self):
         if self._connected:
             self._running.clear()
             Logger.info("TelemetryConnection: closing connection")
-            self._sample_timer.join()
+            if self._sample_timer:
```
brentpicasso: I caught an exception where we were closing and _sample_timer was None, so this guards against an exception.

autosportlabs/telemetry/telemetryconnection.py
```diff
@@ -319,7 +324,7 @@ def handle_connect(self):
     def handle_expt(self):
         # Something really bad happened if we're here
         Logger.error("TelemetryConnection: handle_expt, closing connection")
-        self._update_status("ok", "Unknown error, disconnected from RaceCapture/Live", self.STATUS_DISCONNECTED)
+        self._update_status("error", "Unknown error, disconnected from RaceCapture/Live", self.STATUS_DISCONNECTED)
```
brentpicasso: thought this should be "error" ?

ryandoherty: :+1:

toolbarview.py
```diff
@@ -99,8 +109,11 @@ def on_rc_rx(self, value):
...
 
-    def on_tele_status(self, connected):
+    def on_tele_status(self, status):
         if not self.teleStatus:
             self.teleStatus = kvFind(self, 'rcid', 'teleStatus')
-        self.teleStatus.color = self.txOnColor if connected else self.txOffColor
+        try:        
+            self.teleStatus.color = self.telemetry_color[status]
+        except:
+            pass
```
ryandoherty: We should at least log something here if there's an error, I've torn my hair out debugging things in this code because errors were silently swallowed.

ryandoherty:
Looks good, simple and effective :) Only one thing that probably should be changed is if you disable telemetry after authentication fails (bad device id), the icon color stays red. Probably should change back to the idle color.

toolbarview.py
brentpicasso: +1

brentpicasso:
Good idea. checking if we get a handy notification that the preference changed.

ryandoherty:
Yup, we do get a notification, `on_config_changed` in main.py. Now that I read that function name again, it should probably be changed to `on_preferences_change` :\

brentpicasso:
Ok, just fixed that, was quick.

PR #79, opened 2015-08-28
Add Sentry support

ryandoherty:
Adds basic support for sending crash logs to Sentry.

If a .sentry config file exists locally, it will create a client and send all uncaught exceptions to it.

### DIFF
--- a/main.py
+++ b/main.py
@@ -281,7 +281,6 @@ class RaceCaptureApp(App):
         self._telemetry_connection.telemetry_enabled = False
 
     def showMainView(self, view_name):
-        error = 1/0
         try:
             view = self.mainViews.get(view_name)
             if not view:
--- a/main.py
+++ b/main.py
@@ -11,6 +11,8 @@ if __name__ == '__main__':
     import logging
     import argparse
     import kivy
+    import os
+    import traceback
     from kivy.properties import AliasProperty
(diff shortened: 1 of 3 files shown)


main.py
```diff
@@ -272,6 +281,7 @@ def on_stop(self):
         self._telemetry_connection.telemetry_enabled = False
 
     def showMainView(self, view_name):
+        error = 1/0
```
brentpicasso: should probably remove this.

ryandoherty: LOL oops

main.py
```diff
@@ -486,4 +496,12 @@ def _on_preferences_change(self, menu, config, section, key, value):
                 self._telemetry_connection.telemetry_enabled = False
 
 if __name__ == '__main__':
-    RaceCaptureApp().run()
+    try:
+        RaceCaptureApp().run()
```
brentpicasso: I was going to research if we should re-run the app upon crash to prevent it from exiting completely. I think maybe there are some kivy docs that talk about that.

ryandoherty: Let's do that in a separate PR, this is only for sending crash reports.

brentpicasso: Agreed. Looks good. Shipit

Brent Picasso
Autosport Labs

PR #80, opened 2015-08-28
Updates to buildozer to add sentry file and ignore unnecessary directories when building

ryandoherty:


brentpicasso:
Nice. verified this make a good APK?

PR #81, opened 2015-08-28
Added ExceptionHander and integrated it with Sentry

brentpicasso:


main.py
```diff
@@ -494,13 +495,25 @@ def _on_preferences_change(self, menu, config, section, key, value):
...
 if __name__ == '__main__':
+    ExceptionManager.add_handler(CrashHandler())
     try:
         RaceCaptureApp().run()
     except:
         if 'sentry_client' in globals():
-            ident = sentry_client.captureException()
-            Logger.error("Exception caught; reference is %s" % ident)
+            ident = sentry_client.captureException(stack=True)
+            Logger.error("Main: crash caught: Reference is %s" % ident)
```
stieg: Perhaps do `"%s: Crash caught!  Reference is %s" % (__FILE__, ident)`

Not sure if __FILE__ gets populated in python, but something like that should tell you the file name.

stieg:
Minor comment.  Consider  it.  Then :shipit:

main.py
ryandoherty: Sentry gives us a stack trace with the file(s), so I don't think we need to do it ourselves?

brentpicasso: this is a UUID from sentry / raven client. Not sure why we care about it in the app output?

ryandoherty: Useful if someone is testing or using our app and it crashes, we may be able to ask them for that ID which we can then use to query for in Sentry.

brentpicasso: when we can show last 100 log messages that will be useful, agreed.

brentpicasso: @ryandoherty the docs say stack=False by default, so I thought to add it. I'll disable it.

ryandoherty: Huh, I can see a stack trace in Sentry, unless we're talking about 2 different things?

brentpicasso: Well, I tested it with the stack =true. If stack traces were sent with your
original testing then I think we we good.

Brent Picasso
Autosport Labs

ryandoherty: Cool, either way is fine with me :+1:

PR #83, opened 2015-08-30
select preferred screen on startup; remember last dashboard screen used

brentpicasso:
select the preferred and last screen used.  "JUST WORKS"

### DIFF
--- a/autosportlabs/racecapture/views/dashboard/dashboardview.py
+++ b/autosportlabs/racecapture/views/dashboard/dashboardview.py
@@ -11,6 +11,8 @@ from kivy.uix.screenmanager import *
 from autosportlabs.racecapture.views.dashboard.widgets.tachometer import Tachometer
 from utils import kvFind, kvFindClass
 from autosportlabs.racecapture.views.dashboard.widgets.gauge import Gauge
+from kivy.logger import Logger
+from kivy.clock import Clock
 
 DASHBOARD_VIEW_KV = 'autosportlabs/racecapture/views/dashboard/dashboardview.kv'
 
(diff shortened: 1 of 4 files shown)


main.py
```diff
@@ -395,15 +395,27 @@ def build(self):
...
             Clock.schedule_once(lambda dt: self.first_time_setup(), 0.5)
 
+    def show_startup_view(self):
+        settings_to_view = {'Home Page':'home',
+                            'Dashboard':'dash',
+                            'Analysis': 'analysis',
+                            'Configuration': 'config' }
+        try:
+            view_pref = self.settings.userPrefs.get_pref('preferences', 'startup_screen')
+            view = settings_to_view[view_pref]
```
ryandoherty: We should handle the case here of `get_pref` returning None or False if the setting hasn't been set yet or the prefs file is gone.

main.py
```diff
@@ -395,15 +395,27 @@ def build(self):
...
+            self.showMainView(view)
+        except:
+            Logger.error("RaceCaptureApp: could not show main view " + str(view_pref))
```
ryandoherty: For exceptions, should we log the stack trace or more info as to why the exception occurred? Also I'm wondering if we should start pushing any caught exceptions to Sentry so we can see if there's pieces of the app that aren't working for users.

brentpicasso: I kept it brief b/c the likely case is bad dictionary key. 

Yes, a central handler might be good for non-benign errors!

main.py
brentpicasso: I think we're ok b/c  preferences automatically handles default values.

autosportlabs/racecapture/views/dashboard/dashboardview.py
```diff
@@ -79,12 +81,23 @@ def init_view(self):
...
+
+    def _show_screen(self, screen):
+        self._screen_mgr.current = screen
+        self._settings.userPrefs.set_pref('preferences', 'last_dash_screen', screen)
+
+    def _show_last_view(self):
+        try:
+            last_screen_name = self._settings.userPrefs.get_pref('preferences', 'last_dash_screen')
+            self._screen_mgr.current = last_screen_name
+        except:
```
ryandoherty: Should we do the same thing we did in main.py and remove the try/except and let the userPrefs return the default if the preference isn't found?

brentpicasso: yup, tested and letting the crash handler pick it up works fine.

ryandoherty:
:shipit:

PR #84, opened 2015-09-02
Various cleanups

brentpicasso:
Ensure whitespace in telemetry device ID is stripped out
Remove preferences for features not yet implemented
Prevent labels and values from wrapping

### DIFF
--- a/autosportlabs/racecapture/views/configuration/rcp/telemetryconfigview.py
+++ b/autosportlabs/racecapture/views/configuration/rcp/telemetryconfigview.py
@@ -29,7 +29,7 @@ class TelemetryConfigView(BaseConfigView):
         
     def on_device_id(self, instance, value):
         if self.connectivityConfig:
-            self.connectivityConfig.telemetryConfig.deviceId = value
+            self.connectivityConfig.telemetryConfig.deviceId = strip_whitespace(value)
             self.connectivityConfig.stale = True
             self.dispatch('on_modified')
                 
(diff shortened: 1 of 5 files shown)


autosportlabs/racecapture/views/configuration/rcp/telemetryconfigview.py
```diff
@@ -29,7 +29,7 @@ def __init__(self, **kwargs):
         
     def on_device_id(self, instance, value):
         if self.connectivityConfig:
-            self.connectivityConfig.telemetryConfig.deviceId = value
+            self.connectivityConfig.telemetryConfig.deviceId = strip_whitespace(value)
```
ryandoherty: I think just value.strip() will do what you want. From what I can find in the Python docs strip() will remove newlines.

brentpicasso: strip() only removes whitespace from either end of the string, not within the string. 

strip() docs "Return a copy of the string with leading and trailing characters removed. If chars is omitted or None, whitespace characters are removed. If given and not None, chars must be a string; the characters in the string will be stripped from the both ends of the string this method is called on."

ryandoherty: Ah, I see, misread it :)

utils.py
```diff
@@ -19,6 +20,9 @@
 from kivy import platform
 from datetime import datetime
 
+def strip_whitespace(value):
+    return value.replace('\n','').replace('\r','').strip() if value else ''
```
ryandoherty: the `if value` seems odd, why would someone call strip_whitespace and give it None or False? If they typo a variable name Python will throw an error. If for some reason the variable is None or False it would be better to have Python throw an error than return ''.

brentpicasso: Fair enough. was trying to make it more robust. will remove and let higher level code handle these cases.

ryandoherty:
:shipit:

PR #86, opened 2015-09-02
Added TextWidget for multiline text

ryandoherty:
Basic TextWidget for text that should wrap instead of be trimmed if too long.


brentpicasso:
would be good to put fieldLabel in uix as well, but that can come in as a later refactoring. (there are several things that should move out of the root folder).

:shipit:

PR #87, opened 2015-09-02
Fix alert/warning ranges not working if min or max is 0

ryandoherty:
Python is loosely typed, so a min of 0 is False, so the first conditional would fail.


brentpicasso:
:shipit:

PR #89, opened 2015-09-09
reformatted settings view

brentpicasso:
Adjusted font size and colors to make settings fields stand out better
re-used existing CenteredBubble for error message handling

### DIFF
--- a/autosportlabs/racecapture/views/configuration/rcp/telemetryconfigview.py
+++ b/autosportlabs/racecapture/views/configuration/rcp/telemetryconfigview.py
@@ -4,15 +4,17 @@ from kivy.app import Builder
 from kivy.logger import Logger
 from kivy.uix.boxlayout import BoxLayout
 import re
-
-
-from settingsview import SettingsView, SettingsTextField, SettingsSwitch
-from autosportlabs.widgets.separator import HLineSeparator
-from valuefield import ValueField
-from utils import *
 from autosportlabs.racecapture.views.configuration.baseconfigview import BaseConfigView
+from autosportlabs.uix.toast.kivytoast import toast
+from autosportlabs.racecapture.views.popup.centeredbubble import CenteredBubble, WarnLabel
+from autosportlabs.widgets.separator import HLineSeparator
+from settingsview import SettingsView, SettingsTextField, SettingsSwitch
+from valuefield import ValueField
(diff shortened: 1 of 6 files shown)


autosportlabs/racecapture/views/configuration/rcp/telemetryconfigview.py
```diff
@@ -29,20 +31,26 @@ def __init__(self, **kwargs):
...
             if len(value) > 0:
                 if self.validate_device_id(value):
-                    kvFind(self, 'rcid', 'deviceId').setValue(value)
+                    instance.setValue(value)
                     self.connectivityConfig.telemetryConfig.deviceId = value
                     self.connectivityConfig.stale = True
                     self.dispatch('on_modified')
                 else:
-                    self.ids.device_id.ids.error.text = '[color=#E50000]Id should contain only numbers and letters.[/color]'
+                    warn = CenteredBubble()
```
ryandoherty: I think this will pop up every time the user types a character if one of the preceding characters is invalid. Is that right?

brentpicasso: Correct, b/c it would still be invalid. The user experience seems ok.

ryandoherty: What I meant to say is a new CenteredBubble is being created on each keystroke when instead the code should check to see if the error message is already displayed and if so, reset the timer that hides it after N seconds.

brentpicasso: Yes, it does actually. It's wasteful but relatively harmless and is transparent to the user. Will consider alternative

brentpicasso: Ok, moved it centrally within settingsview, only create it once and quickly dismiss it if warning condition clears.

ryandoherty:
Looks good, minor bug where if you switch to a different screen the error message still shows up:

![screenshot]
(http://i.imgur.com/bJHQCRI.png)

brentpicasso:
Because its attached to the top window. Should still auto dismiss though.
Should see if it's possible to attach it to the parent widget like we do
for channel customization popup.

Brent Picasso
Autosport Labs

PR #90, opened 2015-09-14
Stop telemetry when RCP disconnects

ryandoherty:
Works over BT & USB. Added a sort-of real 'disconnected' event to the RCP api object to make this easier.

### DIFF
--- a/autosportlabs/racecapture/api/rcpapi.py
+++ b/autosportlabs/racecapture/api/rcpapi.py
@@ -72,6 +72,8 @@ class RcpApi:
         self._enable_autodetect = Event()
         self._enable_autodetect.set()
 
+        self._disconnect_callback = kwargs.get('on_disconnect') if 'on_disconnect' in kwargs else None
+
     def enable_autorecover(self):
         Logger.info("RCPAPI: Enabling auto recover")
         self._enable_autodetect.set()
@@ -81,6 +83,9 @@ class RcpApi:
         self._enable_autodetect.clear()
         
     def recover_connection(self):
+        if self._disconnect_callback:
+            self._disconnect_callback()
+
         if self._enable_autodetect.is_set():
             Logger.info("RCPAPI: attempting to recover connection")
             self.run_auto_detect()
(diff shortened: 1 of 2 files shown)


autosportlabs/racecapture/api/rcpapi.py
```diff
@@ -72,6 +72,8 @@ def __init__(self, **kwargs):
         self._enable_autodetect = Event()
         self._enable_autodetect.set()
 
+        self._disconnect_callback = kwargs.get('on_disconnect') if 'on_disconnect' in kwargs else None
```
brentpicasso: I think you'll get the desired effect if you just do self._disconnect_callback = kwargs.get('on_disconnect') - 

if the parameter isn't present the .get() returns None

main.py
```diff
@@ -338,7 +338,7 @@ def build_preferences_view(self):
         preferences_view = PreferencesView(name='preferences', settings=self.settings, base_dir=self.base_dir)
         preferences_view.settings_view.bind(on_config_change=self._on_preferences_change)
         return preferences_view
-    
+    2
```
brentpicasso: syntax error?

main.py
```diff
@@ -426,7 +426,11 @@ def rc_detect_win(self, version):
         if version.is_compatible_version():
             self.showStatus("{} v{}.{}.{}".format(version.friendlyName, version.major, version.minor, version.bugfix), False)
             self.dataBusPump.startDataPump(self._databus, self._rc_api)
-            self._status_pump.start(self._rc_api)        
+            self._status_pump.start(self._rc_api)
+
+            if self.settings.userPrefs.get_pref('preferences', 'send_telemetry') == "1" and self._telemetry_connection:
```
brentpicasso: just refreshed myself on bool() parsing in python. Would be nice to have a helper function in userPrefs to do something like:

userPrefs.is_pref_enabled(....  and would return a bool()

But not attached to it for this PR.

autosportlabs/racecapture/api/rcpapi.py
ryandoherty: :+1:

main.py
ryandoherty: Weird, no idea how that got there.

main.py
ryandoherty: :+1:

brentpicasso:
:sake:

PR #91, opened 2015-09-24
Increase brightness of android checkboxes with custom theme file

ryandoherty:


build_release_apk.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -e
 rm -rf bin
+cp install/defaulttheme-0.png .buildozer/android/platform/python-for-android/dist/racecapture/private/lib/python2.7/site-packages/kivy/data/images/defaulttheme-0.png
```
brentpicasso: wow, right into the .buildozer folder. Seems extreme but awesome. Is this primarily b/c they don't have a good way to do this officially?

brentpicasso:
Think we need to worry about this for windows?

build_release_apk.sh
ryandoherty: Yeah, I had tried to do it the (semi) 'official' way in Kivy by calling some kivy functions to load a different theme file, but it never worked and there's little to no documentation on how custom theming works. This is relatively safe b/c if for some reason the file name changes with a Kivy upgrade, you just get the Kivy theme instead.

ryandoherty:
Windows seems to be ok with the default theme because of either higher contrast monitors or their color profiles.

brentpicasso:
Great! if tested and looks good on android then update changelog and then :shipit: Also, maybe update version in main.py to 1.3.10 as well.

ryandoherty:
Yup, tested on android and OSX to verify changes didn't break the OSX build/theme. Will update changelog.

PR #92, opened 2015-09-25
Faster trackmap update

ryandoherty:
- Speed up firstrun track map downloading by ~7x
- Speed up refreshing current tracks by ~2x
- PEP8 compliance
- Lots of code cleanup and comments

I started with just speeding up the download/refresh progress, then I got caught up in cleaning up one thing, then one thing led to another and I rewrote nearly all of it. 

I removed the Venue class, it wasn't being used outside of TrackManager, and only in a limited way. Was simpler to just use the JSON data from RCL instead of creating Venue objects that were then immediately discarded.

I wouldn't call this 100% perfect, but I gotta stop polishing at some point :P

### DIFF
--- a/autosportlabs/racecapture/config/rcpconfig.py
+++ b/autosportlabs/racecapture/config/rcpconfig.py
@@ -669,14 +669,14 @@ class Track(object):
     @classmethod
     def fromTrackMap(cls, trackMap):
         t = Track()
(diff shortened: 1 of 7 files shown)


brentpicasso:
closing this PR as I'm carrying PR forward with additional fixes.

PR #95, opened 2015-10-06
P/brentpicasso/get log

brentpicasso:
System status view now lets you access the recent application log. A button allows the user to copy the last 100 lines of the log to the system clipboard.

app_log

### DIFF
--- a/autosportlabs/racecapture/views/status/statusview.py
+++ b/autosportlabs/racecapture/views/status/statusview.py
@@ -10,8 +10,7 @@ from kivy.properties import ObjectProperty, StringProperty, NumericProperty
 from datetime import timedelta
 from utils import *
 from fieldlabel import FieldLabel
-from kivy.logger import LoggerHistory
-from kivy.core.clipboard import Clipboard
+from kivy.logger import LoggerHistory, Logger
 from autosportlabs.racecapture.theme.color import ColorScheme
 from autosportlabs.uix.toast.kivytoast import toast
 
@@ -55,7 +54,7 @@ class StatusView(Screen):
 
     #Used for building the left side menu
     _menu_keys = {
-        "applog": "Application",
+        "app": "Application",
         "system": "RaceCapture",
         "GPS": "GPS",
         "cell": "Cellular",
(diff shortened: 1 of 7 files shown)


autosportlabs/racecapture/views/status/statusview.py
```diff
@@ -10,11 +10,14 @@
...
 from fieldlabel import FieldLabel
+from kivy.logger import LoggerHistory, Logger
+from autosportlabs.racecapture.theme.color import ColorScheme
+from autosportlabs.uix.toast.kivytoast import toast
 
 STATUS_KV_FILE = 'autosportlabs/racecapture/views/status/statusview.kv'
 
-RAW_STATUS_BGCOLOR_1 = [0  , 0  , 0  , 1.0]
-RAW_STATUS_BGCOLOR_2 = [0.10, 0.10, 0.10, 1.0]
+RAW_STATUS_BGCOLOR_1 = ColorScheme.get_background()
```
stieg: Good.

autosportlabs/racecapture/views/status/statusview.py
```diff
@@ -128,29 +132,30 @@ def __init__(self, track_manager, status_pump, **kwargs):
...
+        self._menu_node.select_node(self._menu_node.root.nodes[0])
         
     def _build_menu(self):
         if self._menu_built:
             return
 
-        default_node = None
-
-        for item, config in self.status.iteritems():
+        for item in self.status.iterkeys():
```
stieg: Need to be careful that the status dictionary won't change on us when iterating the keys.  Else we will get an exception

stieg:
Does it allow you to set the logging level as well?

stieg:
Minor nits.  Else good to go.

autosportlabs/racecapture/views/status/statusview.py
brentpicasso: good point. near zero chance of this happening in this case, however b/c how the notifications propagate.

brentpicasso:
Setting the logging level would be a good future enhancement.

autosportlabs/racecapture/views/status/statusview.py
stieg: I've been seeing errors related to this come in on Sentry so it may not be a close to 0 as you might think... but your call here.

stieg: Well... not this code specifically... but in general.  You know what I mean.

PR #778, opened 2015-12-29
Basic analysis

brentpicasso:
Basic analysis for beta. this one is a doozy.


ryandoherty:
![shocked]
(http://i1212.photobucket.com/albums/cc446/inshirtsleeves/shocked.gif)

autosportlabs/help/helpmanager.py
```diff
@@ -0,0 +1,114 @@
...
+        except Exception as e:
+            Logger.critical('HelpInfo: Failed to show help popup: {} {}'.format(e, traceback.format_exc()))
+
+    def on_ok(self):
+        HelpInfo.settings.userPrefs.set_pref('help', self._key, False)
+        self.parent.parent.dismiss()
+        
+    
+    
+
```
ryandoherty: Lots of blank lines here.

autosportlabs/help/helpmanager.py
```diff
@@ -0,0 +1,114 @@
...
+from kivy.metrics import sp
+from kivy.app import Builder
+import json
+import os
+import traceback
+
+__all__ = ('check_help_popup_popup')
+
+HELP_INFO_LAYOUT='''
+<HelpInfo>:
```
ryandoherty: Any particular reason for embedding this instead of a separate file like most other layouts/views?

autosportlabs/help/helpmanager.py
```diff
@@ -0,0 +1,114 @@
...
+                if helptext:
+                    help_popup = HelpBubble()
+                    help_popup.arrow_pos = arrow_pos
+                    content = HelpInfo(key)
+                    content.title_text = helptext['title']
+                    content.help_text = helptext['text']
+                    help_popup.add_widget(content)
+                    help_popup.auto_dismiss_timeout(HelpInfo.HELP_POPUP_TIMEOUT)
+                    help_popup.size = HelpInfo.HELP_POPUP_SIZE
+                    help_popup.size_hint = (None,None)
```
ryandoherty: All these assignments to `help_popup` make me think they could just be passed into the constructor. Thoughts?

autosportlabs/help/helpmanager.py
brentpicasso: This was an example adopted from elsewhere and it's used sporadically in the app elsewhere. 

I'm thinking maybe it could be a better option than separate files since we have to specify the absolute-ish path otherwise. 

Side note, I'm trying to find a way to lazy load the layout only when the class loads, to speed up class load time. I'm putting the builder.load() as a class initializer.

autosportlabs/help/helpmanager.py
brentpicasso: indeed

autosportlabs/racecapture/datastore/datastore.py
```diff
@@ -204,37 +219,78 @@ def group(self, filterchain):
...
+    @property
+    def is_open(self):
+        return self._isopen
+
+    @property
+    def channel_list(self):
+        return self._channels[:]
+
+    def get_channel(self, name): 
+        channel =  [c for c in self._channels if name in c.name]
```
ryandoherty: :+1:

autosportlabs/help/helpmanager.py
brentpicasso: It's a static method, so designed to be called with a unique key and widget for each help message.

brentpicasso:
^^ is watching for more feedback  http://cdn2.teen.com/wp-content/uploads/2014/03/lies-kids-tell-parents-10.gif

ryandoherty:
Just did a very rough skim of the code. There is about a 1000:1 ratio of comments to code, which could be improved. Not the normal comments that just state the obvious, but actual comments as to how to use a class, why it exists and to help describe potentially confusing sections (notably the data import/massaging that occurs). 

The widgets for the line chart, map, channel selection and others could use some comments on how the class can be used because they will be reused a lot throughout the analysis sections.

Glad to see tests were added! The part about requiring an order to the tests has a code smell to me. There are setup and teardown methods that can be used to avoid forcing test cases to run first and last. I can see that importing data could take a while, maybe we can import small chunks of data per test case? Or a smaller dataset so the tests can still run fast.

Liking how this new code is more Pythonic and less C ;) :+1: 

Long-term we will need a sane way of setting up the database and migrating data between schemas as we add new functionality. Not require for beta or GA, but something to start thinking about. Python based database migration tools exist and many are very mature. 

I'll review it more over the next few days and add more comments.

brentpicasso:
That's good feedback. For 1.4.1 we'll need a table to store the database version so we know how to do an upgrade as necessary.

autosportlabs/help/helpmanager.py
ryandoherty: Yeah, I'm kind leaning towards this too after having to dig through files to find layouts, etc.

autosportlabs/help/helpmanager.py
ryandoherty: The method is static, but it creates a new instance of HelpBubble each time it's called. Minor nitpick, would just simplify the use of HelpBubble. Or make HelpBubble static too and have an `update` method?

autosportlabs/racecapture/datastore/datastore.py
```diff
@@ -126,7 +129,13 @@ def fetch_records(self, count):
 
         return zip(*zlist)
 
-
+class Session(object):
+    def __init__(self, ses_id, name, notes = '', date = None):
+        self.ses_id = ses_id
```
ryandoherty: Minor naming comment, thinking self.id or self.session_id would be clearer.

autosportlabs/racecapture/datastore/datastore.py
```diff
@@ -204,37 +219,78 @@ def group(self, filterchain):
...
         self.sample_rate = sample_rate
 
+    def __str__(self):
+        return self.name
 
 class DataStore(object):
+    EXTRA_INDEX_CHANNELS = ["LapCount"]    
     val_filters = ['lt', 'gt', 'eq', 'lt_eq', 'gt_eq']
-    def __init__(self, name=':memory:'):
+    def __init__(self):
```
ryandoherty: Why is the database path is passed into `open_db` instead of the constructor?

autosportlabs/racecapture/datastore/datastore.py
```diff
@@ -329,7 +398,7 @@ def _get_last_table_id(self, table_name):
         #print "Last datapoint ID =", dp_id
         return dl_id
 
-    def _insert_record(self, record, channels, session_id):
+    def _insert_record(self, record, channels, session_id, warnings = None):
```
ryandoherty: What's `warnings` used for?

autosportlabs/racecapture/datastore/datastore.py
```diff
@@ -401,7 +470,7 @@ def _extrap_datapoints(self, datapoints):
         return extrap_list
 
 
-    def _desparsified_data_generator(self, data_file):
+    def _desparsified_data_generator(self, data_file, warnings=None, progress_cb=None):
```
ryandoherty: \o/ for progress callbacks/bars!

autosportlabs/racecapture/datastore/datastore.py
```diff
@@ -456,13 +538,22 @@ def _desparsified_data_generator(self, data_file):
...
+            if work_list_len == 0:
                 work_list = [[] for x in channels]
                 yield_list = [[] for x in channels]
+            
+            if channels_len > work_list_len and current_line > 0:
+                warn_msg = 'Unexpected channel count in line {}. Expected {}, got {}'.format(current_line, work_list_len, channels_len)
+                if warnings:
+                    warnings.append((line, warn_msg))
+                Logger.warn("DataStore: {}".format(warn_msg))
+                continue
```
ryandoherty: What happens if this situation occurs? Is it common?

autosportlabs/racecapture/datastore/datastore.py
```diff
@@ -518,51 +631,88 @@ def _create_session(self, name, notes=''):
...
-        base_sql = "SELECT {} from datapoint ORDER BY {} DESC LIMIT 1;".format(channel, channel)
-        c.execute(base_sql)
+        base_sql = "SELECT AVG(Latitude), AVG(Longitude) from datapoint"
+        
+        if type(sessions) == list and len(sessions) > 0:
+            base_sql += " JOIN sample ON datapoint.sample_id=sample.id WHERE "            
+            ses_filters = []
+            for s in sessions:
+                ses_filters.append('(sample.session_id = {}'.format(s))
+            base_sql += '     OR '.join(ses_filters)
```
ryandoherty: I think you can use an `IN()` here instead.

autosportlabs/racecapture/datastore/datastore.py
```diff
@@ -518,51 +631,88 @@ def _create_session(self, name, notes=''):
...
+        return (lat_average, lon_average)
+                
+    def _session_select_clause(self, sessions=None):
+        sql = ''
+        if type(sessions) == list and len(sessions) > 0:
+            sql += " JOIN sample ON datapoint.sample_id=sample.id WHERE "            
+            ses_filters = []
+            for s in sessions:
+                ses_filters.append('sample.session_id = {}'.format(s))
+            sql += '     OR '.join(ses_filters)
```
ryandoherty: Same about `IN()`.

autosportlabs/racecapture/datastore/datastore.py
```diff
@@ -616,15 +770,23 @@ def import_datalog(self, path, name, notes='', progress_listener=None):
...
+        self._handle_data(dl, headers, ses_id, warnings, progress_cb)
+        return ses_id
 
-    def query(self, channels=[], data_filter=None):
+    def query(self, sessions=[], channels=[], data_filter=None, distinct_records=False):
         #Build our select statement
         sel_st  = 'SELECT '
 
+        if distinct_records:
+            sel_st += 'DISTINCT '
```
ryandoherty: Man oh man all this raw query building makes me think we need an ORM to handle this.

autosportlabs/racecapture/datastore/datastore.py
```diff
@@ -603,7 +752,12 @@ def get_channel_smoothing(self, channel):
             return res[0]
 
     @timing
-    def import_datalog(self, path, name, notes='', progress_listener=None):
+    def import_datalog(self, path, name, notes='', progress_cb=None):
+        
+        warnings = []
+        if not self._isopen:
+            raise Exception("Datastore is not open")
```
ryandoherty: Might be better to create a new Exception class that's specific to this scenario so other code can catch it if needed instead of having to catch all Exceptions.

autosportlabs/racecapture/datastore/datastore.py
```diff
@@ -616,15 +770,23 @@ def import_datalog(self, path, name, notes='', progress_listener=None):
...
 
+        if distinct_records:
+            sel_st += 'DISTINCT '
+
         columns = []
         joins = []
 
+        #make sure that the sessions list exists
+        if type(sessions) != list or len(sessions) == 0:
+            raise Exception("Must provide a list of sessions to query!")
```
ryandoherty: Same RE: Exceptions.

autosportlabs/racecapture/datastore/datastore.py
```diff
@@ -648,15 +814,29 @@ def query(self, channels=[], data_filter=None):
         #Add our joins
         sel_st += 'JOIN datapoint ON datapoint.sample_id=sample.id\n'
 
-        #Add our filter
-        sel_st += 'WHERE '
-
         if not data_filter == None:
```
ryandoherty: Double negative. (I know this isn't part of your changes, but hey while you're here :P )

ryandoherty:
Also RE: database structure, can we extract the SQL used to generate the tables into a separate file for now? Would make it a lot easier to understand the DDL.

And can the table names be pluralized (ones that aren't), fairly standard for table names.

autosportlabs/racecapture/settings/prefs.py
```diff
@@ -71,6 +72,9 @@ def set_gauge_config(self, gauge_id, channel):
         self._prefs_dict["gauge_settings"][gauge_id] = channel
         self._schedule_save()
 
+    def get_datastore_location(self):
```
ryandoherty: Could this and other `get`s be properties instead?

autosportlabs/racecapture/tracks/trackmanager.py
```diff
@@ -166,14 +167,15 @@ def find_track_by_short_id(self, short_id):
...
+    def find_nearby_track(self, point, searchRadius = TRACK_DEFAULT_SEARCH_RADIUS_METERS, searchBearing = TRACK_DEFAULT_SEARCH_BEARING_DEGREES):
+        radius = point.metersToDegrees(searchRadius, searchBearing)
+        for trackId in self.tracks.keys():
+            track = self.tracks[trackId]
+            trackCenter = track.centerpoint
+            if trackCenter and trackCenter.withinCircle(point, radius):
                 return track
         return None
-        
+
```
ryandoherty: Whitespace!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

autosportlabs/racecapture/views/analysis/addstreamview.kv
```diff
@@ -0,0 +1,136 @@
...
+	orientation: 'vertical'
+	ScreenManager:
+		id: screens
+		AddStreamSelectView:
+			id: streamSelectScreen
+			name: 'stream_select'
+        CloudConnectView:
+            id: cloudConnectScreen
+            name: 'cloud_connect'
+        WirelessConnectView:
```
ryandoherty: At this point if we're not going to ship wireless and cloud downloading with GA, let's hide these icons. No more teasing users :)

autosportlabs/racecapture/views/analysis/addstreamview.py
```diff
@@ -0,0 +1,255 @@
...
+from iconbutton import IconButton
+from fieldlabel import FieldLabel
+
+Builder.load_file('autosportlabs/racecapture/views/analysis/addstreamview.kv')
+
+class AddStreamView(BoxLayout):
+    def __init__(self, **kwargs):
+        super(AddStreamView, self).__init__(**kwargs)
+        settings = kwargs.get('settings')
+        datastore = kwargs.get('datastore')
```
ryandoherty: If an argument is required, I think it should be explicit and a normal argument.

autosportlabs/racecapture/views/analysis/addstreamview.py
```diff
@@ -0,0 +1,255 @@
...
+        self.dispatch('on_connect_stream_complete', session_id)
+        
+    def on_connect_file_stream(self, *args):
+        self.ids.screens.current = 'file_connect'        
+
+    def on_connect_cloud_stream(self, *args):
+        self.ids.screens.current = 'cloud_connect'
+        
+    def on_connect_wireless_stream(self, *args):
+        self.ids.screens.current = 'wireless_connect'
```
ryandoherty: Are any of these event handlers supposed to do anything or are they just stubbed out?

ryandoherty: Also, I think the event handlers for this are backwards or not needed. This class is binding event handlers to events fired by the view, correct? If so, this class doesn't need to fire its own events.

autosportlabs/racecapture/views/analysis/addstreamview.py
```diff
@@ -0,0 +1,255 @@
...
+        self.dispatch('on_connect_file')
+
+    def on_connect_cloud(self, *args):
+        pass
+    
+    def on_connect_wireless(self, *args):
+        pass
+        
+    def on_connect_file(self, *args):
+        pass
```
ryandoherty: Same questions RE events.

autosportlabs/racecapture/views/analysis/addstreamview.py
```diff
@@ -0,0 +1,255 @@
...
+            
+            if os.path.isfile(dstore_path):
+                self.datastore.open_db(dstore_path)
+            else:
+                self.datastore.new(dstore_path)
+
+        Logger.info("LogImportWidget: loading log: {}".format(self.ids.log_path.text))
+
+        if not session_name or len(session_name) == 0:
+            alertPopup('Error', 'A session name must be specified')
```
ryandoherty: Inline errors would be nicer here instead of popups.

autosportlabs/racecapture/views/analysis/addstreamview.py
```diff
@@ -0,0 +1,255 @@
...
+        
+        dstore_path = self.settings.userPrefs.get_datastore_location()
+
+        if not logpath:
+            alertPopup("No Log Specified",
+                      "Please select a log file to import")
+            return
+
+        if not os.path.isfile(logpath):
+            alertPopup("Invalid log specified",
```
ryandoherty: Inline errors would be nicer here instead of popups.

autosportlabs/racecapture/views/analysis/analysisview.py
```diff
@@ -1,35 +1,180 @@
...
+from autosportlabs.racecapture.views.file.savedialogview import SaveDialog
+from autosportlabs.racecapture.views.util.alertview import alertPopup
+from autosportlabs.uix.color.colorsequence import ColorSequence
+from autosportlabs.racecapture.theme.color import ColorScheme
+from autosportlabs.help.helpmanager import HelpInfo
+import traceback
 
 ANALYSIS_VIEW_KV = 'autosportlabs/racecapture/views/analysis/analysisview.kv'
-
+
```
ryandoherty: Whitespace!

autosportlabs/racecapture/views/analysis/analysisview.py
```diff
@@ -1,35 +1,180 @@
...
+        source_key = str(source_ref)
+        if selected:
+            self.ids.mainchart.add_lap(source_ref)
+            self.ids.channelvalues.add_lap(source_ref)
+            map_path_color = self._color_sequence.get_color(source_key)
+            self.ids.analysismap.add_reference_mark(source_key, map_path_color)
+            cache = self._datastore.get_location_data(source_ref)
+            self._sync_analysis_map(source_ref.session)
+            self.ids.analysismap.add_map_path(source_ref, cache, map_path_color)
+          #  self.ids.analysismap.add_heat_values('TPS', source_ref)
```
ryandoherty: If not needed, remove.

autosportlabs/racecapture/views/analysis/analysisview.py
```diff
@@ -1,35 +1,180 @@
...
+        self.ids.channelvalues.color_sequence = self._color_sequence
+        self.ids.mainchart.color_sequence = self._color_sequence
+        self.stream_connecting = False
         self.init_view()
+
+    def on_sessions(self, instance, value):
+        self.ids.channelvalues.sessions = value
+        
+    def lap_selected(self, instance, source_ref, selected):
+        source_key = str(source_ref)
```
ryandoherty: Why the casting here? What type is `source_ref` and what does it mean?

autosportlabs/racecapture/views/analysis/analysisview.py
```diff
@@ -1,35 +1,180 @@
...
+                if best_lap:
+                    best_lap_id = best_lap[1]
+                    Logger.info('AnalysisView: Convenience selected a suggested session {} / lap {}'.format(new_session_id, best_lap_id))
+                    sessions_view.select_lap(new_session_id, best_lap_id, True)
+                    main_chart = self.ids.mainchart
+                    main_chart.select_channels(AnalysisView.SUGGESTED_CHART_CHANNELS)
+                    HelpInfo.help_popup('suggested_lap', main_chart, arrow_pos='left_mid')
+                else:
+                    Logger.warn('AnalysisView: Could not determine best lap for session {}'.format(new_session_id))
+        except Exception as e:
```
ryandoherty: Usually best to catch the types of exceptions you expect, catching all of them can cause weird behavior or make it difficult to debug issues.

autosportlabs/racecapture/views/analysis/analysiswidget.py
```diff
@@ -0,0 +1,153 @@
...
+    
+    def remove_lap(self, lap_ref):
+        try:
+            self.on_lap_removed(lap_ref)
+            del(self.selected_laps[str(lap_ref)])
+        except Exception as e:
+            Logger.error("AnalysisWidget: Error removing remove lap " + str(e))
+
+class ChannelAnalysisWidget(AnalysisWidget):
+    """A widget that can select it's own channels to display
```
ryandoherty: s/it's/its

http://www.its-not-its.info/

autosportlabs/racecapture/views/analysis/channelvaluesview.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def update_reference_mark(self, source, point):
+        channel_data = self.channel_stats.get(str(source))
+        for channel, channel_data in channel_data.iteritems():
+            stats = channel_data.data
+            key = channel + str(source)
+            widget = self._channel_stat_widgets.get(key)
+            #widget.session = str(source.session)
+            #widget.lap = str(source.lap)
+            #widget.channel = channel
```
ryandoherty: If not needed, remove.

autosportlabs/racecapture/views/analysis/analysisview.py
brentpicasso: source_ref is an object of type SourceRef. the string generator was overridden to generate a useful key when needed as a string

autosportlabs/racecapture/datastore/datastore.py
brentpicasso: It's there to handle problems in the logfile. Ran across one logfile with issues.

autosportlabs/racecapture/datastore/datastore.py
brentpicasso: We're going to collect warnings and present them to the user after import is complete.

autosportlabs/help/helpmanager.py
brentpicasso: I think object creation is pretty cheap in python and it seems to follow the kivy pattern better, creating a new instance and then letting kivy manage it.

autosportlabs/racecapture/datastore/datastore.py
brentpicasso: Never found an ORM framework I liked. These kind of frameworks:  "Make the easy things easier, and the hard things impossible"

-but I'm just cynical about them :)

autosportlabs/racecapture/datastore/datastore.py
brentpicasso: yeah, ses_id is used throughout the codebase. did a search. Hmm.

autosportlabs/racecapture/datastore/datastore.py
brentpicasso: I'm guessing so we can close and re-open a different database file so we can manage different data stores (so people can manage different collections of sessions)

autosportlabs/racecapture/datastore/datastore.py
brentpicasso: checked SQLite, I think that would work!

autosportlabs/racecapture/datastore/datastore.py
brentpicasso: fixing.

autosportlabs/racecapture/views/analysis/analysisview.py
brentpicasso: fixed

autosportlabs/racecapture/views/analysis/analysiswidget.py
brentpicasso: fixed

autosportlabs/racecapture/views/analysis/channelvaluesview.py
brentpicasso: fixed

autosportlabs/racecapture/views/analysis/analysisview.py
brentpicasso: fixed!

autosportlabs/racecapture/views/analysis/analysisview.py
brentpicasso: We can let the crash handler catch this so it can be reported. WFM.

autosportlabs/racecapture/views/analysis/channelvaluesview.py
```diff
@@ -0,0 +1,149 @@
...
+    def maxval(self, value):
+        self.value_view.maxval = value
+
+
+class ChannelValuesView(ChannelAnalysisWidget):
+    color_sequence = ObjectProperty(None)
+
+    def __init__(self, **kwargs):
+        super(ChannelValuesView, self).__init__(**kwargs)
+        self.channel_stats={}
```
ryandoherty: Space plzkthxbai.

autosportlabs/racecapture/views/analysis/channelvaluesview.py
```diff
@@ -0,0 +1,149 @@
...
+                view.maxval = channel_data.max
+                self._channel_stat_widgets[key] = view
+                
+        channels_grid.clear_widgets()
+        for key in iter(sorted(self._channel_stat_widgets.iterkeys())):
+            channels_grid.add_widget(self._channel_stat_widgets[key])
+
+    def add_channel(self, channel_data):
+        source_key = str(channel_data.source)
+        channels = self.channel_stats.get(source_key)
```
ryandoherty: Is `self.channel_stats.get(source_key)` returning one or multiple channels? I'm confused how this method works. Does it add one channel or multiple?

autosportlabs/racecapture/views/analysis/sessionbrowser.py
```diff
@@ -0,0 +1,231 @@
...
+        self.notes = notes
+        self.lap_count = 0
+        self.register_event_type('on_delete_session')
+        self.register_event_type('on_edit_session')
+    
+    def append_lap(self, session, lap, laptime):
+        text = '{} :: {}'.format(int(lap), format_laptime(laptime))
+        lapitem = LapItemButton(session=session, text=text, lap=lap, laptime=laptime)
+        self.ids.lap_list.add_widget(lapitem)
+        self.lap_count += 1
```
ryandoherty: Any reason we're manually keeping track of the lap_count?

autosportlabs/racecapture/views/analysis/sessionbrowser.py
```diff
@@ -0,0 +1,231 @@
...
+            self.clear_sessions()
+            sessions = self.datastore.get_sessions()
+            f = Filter().gt('LapCount', 0)
+            session = None
+            for session in sessions:
+                session = self.append_session(ses_id=session.ses_id, name=session.name, notes=session.notes)
+                dataset = self.datastore.query(sessions=[session.ses_id],
+                                        channels=['LapCount', 'LapTime'],
+                                        data_filter=f,
+                                        distinct_records=True)
```
ryandoherty: Instead of querying across all data to get lap count and times, we could do what RCL does. It watches the stream of data coming from RCP and generates a new row in a laps table when the lapcount changes. That way you don't have to query every time to figure out laps. This could be done when you import a log file. You could also compute the min/max/avgs on import too.

autosportlabs/racecapture/views/analysis/sessionbrowser.py
```diff
@@ -0,0 +1,231 @@
...
+        session_editor.session_name = session.name
+        session_editor.session_notes = session.notes
+        popup = editor_popup('Edit Session', session_editor, _on_answer)
+    
+    def delete_session(self, instance, id):
+        try:
+            self.datastore.delete_session(id)
+            Logger.info('SessionBrowser: Session {} deleted'.format(id))
+            self.refresh_session_list()
+        except Exception as e:
```
ryandoherty: Generic error catching comment.

autosportlabs/racecapture/views/analysis/sessioneditorview.py
```diff
@@ -0,0 +1,29 @@
+import os
+from threading import Thread
+import kivy
+from kivy.uix.boxlayout import BoxLayout
+
+kivy.require('1.8.0')
```
ryandoherty: Aren't we on 1.9?

autosportlabs/racecapture/views/channels/channelselectview.py
```diff
@@ -15,6 +17,53 @@ class ChannelItemButton(ListItemButton):
     def __init__(self, **kwargs):
         super(ChannelItemButton, self).__init__(**kwargs)
 
+class ChannelSelectorView(BoxLayout):
+    channels = ListProperty()
+    multi_select = False
+    def __init__(self, **kwargs):
+        super(ChannelSelectorView, self).__init__(**kwargs)
+        self.register_event_type('on_channel_selected')
+        self.multi_select = kwargs.get('multi_select', self.multi_select)
```
ryandoherty: I think if `self.multi_select` is an instance variable it should only be declared inside `__init__` https://docs.python.org/2/tutorial/classes.html#class-and-instance-variables

autosportlabs/racecapture/views/channels/channelselectview.py
```diff
@@ -15,6 +17,53 @@ class ChannelItemButton(ListItemButton):
...
+                    view.trigger_action(duration=0)
+                index += 1
+        
+    def on_channels(self, instance, value):    
+        data = []
+        channel_list = self.ids.channelList
+        for channel in self.channels:
+            data.append({'text': str(channel), 'is_selected': False})
+        
+        args_converter = lambda row_index, rec: {'text': rec['text'], 'size_hint_y': None, 'height': dp(50)}
```
ryandoherty: What's going on here? Can we get a comment or two about what/why this is?

autosportlabs/racecapture/views/channels/channelselectview.py
```diff
@@ -15,6 +17,53 @@ class ChannelItemButton(ListItemButton):
...
+    def on_channel_selected(self, channel):
+        pass
+    
+    def on_select(self, value):
+        try:
+            channels = []
+            for channel in value.selection:
+                channels.append(channel.text)
+            self.dispatch('on_channel_selected', channels)
+        except Exception as e:
```
ryandoherty: Generic exception catching statement.

autosportlabs/racecapture/views/dashboard/comboview.py
```diff
@@ -13,7 +15,21 @@ class ComboView(Screen):
 
     def __init__(self, **kwargs):
         super(ComboView, self).__init__(**kwargs)
+        self.register_event_type('on_tracks_updated')
+        self._databus = kwargs.get('dataBus')
+        self._settings = kwargs.get('settings')
```
ryandoherty: If args are required we should make them normal method arguments.

autosportlabs/racecapture/views/dashboard/comboview.py
```diff
@@ -13,7 +15,21 @@ class ComboView(Screen):
...
     def init_view(self):
+        data_bus = self._databus
+        settings = self._settings
+        
+        gauges = list(kvFindClass(self, Gauge))
+        
+        for gauge in gauges:
+            gauge.settings = settings
+            gauge.data_bus = data_bus        
+        pass
```
ryandoherty: I think `pass` isn't needed or incorrect here.

autosportlabs/racecapture/views/dashboard/widgets/imugauge.py
```diff
@@ -3,16 +3,78 @@
...
-    def initWidgets(self):
-        pass
+    
+    def on_channel_meta(self, channel_metas):
+        self.channel_meta = channel_metas
+                        
+    def on_data_bus(self, instance, value):
+        self._update_channel_binding()
+    
+    def updateColors(self):
```
ryandoherty: Should be `update_colors`

autosportlabs/racecapture/views/util/alertview.py
```diff
@@ -5,9 +5,11 @@
 from kivy.uix.gridlayout import GridLayout
 from kivy.metrics import dp
 from kivy.app import Builder
-from kivy.properties import StringProperty
+from kivy.properties import StringProperty, ObjectProperty
+from kivy.clock import Clock
+from iconbutton import IconButton
 
-__all__ = ('alertPopup, confirmPopup, okPopup')
+__all__ = ('alertPopup, confirmPopup, okPopup, editor_popup')
```
ryandoherty: Just noticed this, what is it used for?

autosportlabs/racecapture/views/util/viewutils.py
```diff
@@ -0,0 +1,25 @@
...
+
+NULL_LAP_TIME = '--:--.---'
+MIN_LAP_TIME=0
+
+def format_laptime(time):
+    if time == None:
+        return NULL_LAP_TIME
+
+    intMinuteValue = int(time)
+    fractionMinuteValue = 60.0 * (time - float(intMinuteValue))
```
ryandoherty: Not PEP8, should be underscored.

autosportlabs/uix/color/colorsequence.py
```diff
@@ -0,0 +1,26 @@
+from kivy.utils import get_color_from_hex as rgb
+
+#DEFAULT_COLOR_SEQUENCE = ['FFFFFF', '003AC1','FF5D00','FFC700','607DC1','FFAE7F','FFE37F','919FC1','FFD6BF','FFF1BF']
+DEFAULT_COLOR_SEQUENCE = ['FFFFFF', '8A00B8', '3366FF', 'F5B800', '8AB800']
+#DEFAULT_COLOR_SEQUENCE = ['FFFFFF', '5DA5DA', 'FAA43A','60BD68', 'F17CB0', 'B2912F', 'B276B2','DECF3F', 'F15854', 'FFFFFF']
+#DEFAULT_COLOR_SEQUENCE =  ['2b908f', '90ee7e', 'f45b5b', '7798BF', 'aaeeee', 'ff0066', 'eeaaee', '55BF3B', 'DF5353', '7798BF', 'aaeeee']
```
ryandoherty: These could be removed if not used anymore.

autosportlabs/uix/track/racetrackview.py
```diff
@@ -8,19 +8,40 @@
...
 from utils import *
 
 Builder.load_file('autosportlabs/uix/track/racetrackview.kv')
         
 class RaceTrackView(BoxLayout):
-    trackmap = None
     def __init__(self, **kwargs):
         super(RaceTrackView, self).__init__(**kwargs)
-
+
```
ryandoherty: Whitespace!

autosportlabs/uix/track/trackmap.py
```diff
@@ -5,56 +5,133 @@
...
+        self.offsetPoint = Point(0,0)
+        self.globalRatio = 0
+        self.heightPadding = 0
+        self.widthPadding = 0
+        
+        self.minXY = Point(-1, -1)
+        self.maxXY = Point(-1, -1)
+    
+        #The trackmap
+        self.mapPoints = []
```
ryandoherty: A few non-PEP8 compliant variables here and above.

resource/help/help_text.json
```diff
@@ -0,0 +1,10 @@
+{
+    "suggested_lap": {
+        "title": "Ready to review data!",
+        "text": "We've selected the best lap and a suggested channel to get you started.\n\nTo add more, touch the gear icon for a list of available channels."
+    },
+    "beta_analysis_welcome":{
+        "title": "Beta Analysis",
+        "text": "To get started, touch the (+) icon to import a log file. Sample logfiles are available from our support forums.\n\nBeta software! Expect significant improvements. Side effects may include needing to re-import your data."    
+    }
+}
```
ryandoherty: Any reason why this text is extracted?

autosportlabs/racecapture/views/analysis/sessionbrowser.py
brentpicasso: we're using it externally  (line 151 of same file) to help vertically size the accordion. I was thinking of just counting the number of child widgets of (self.ids.lap_list) to eliminate tracking of a variable. Will check on that.

brentpicasso: Yup, eliminated this variable.

autosportlabs/racecapture/views/analysis/sessionbrowser.py
brentpicasso: Sounds promising. How about for 1.4.1, when we add a screen for min / max / average / std. deviation?

autosportlabs/racecapture/views/analysis/sessionbrowser.py
brentpicasso: I think we still need the alert popup, but we should maybe re-raise it so the global error handler / sentry can catch and report it?

autosportlabs/racecapture/views/analysis/sessioneditorview.py
brentpicasso: swept through and applied 1.9.0 throughout codebase.

autosportlabs/racecapture/views/channels/channelselectview.py
brentpicasso: :+1:  agree, as well as channels. I've been usually doing this wherever I see it. We have more to fix elsewhere in code too

brentpicasso: nvm on channels, that's a kivy property. left it outside.

fixed otherwise.

autosportlabs/racecapture/views/channels/channelselectview.py
brentpicasso: fixed.

autosportlabs/uix/color/colorsequence.py
brentpicasso: I think the color scheme we have now is adequate, so we can get rid of the other ones.

brentpicasso: fixed.

resource/help/help_text.json
brentpicasso: I wanted to put all help text in one place, and it's a nod towards future localization. 

It's one thing to put human readable text in the .kv file, but it's worse to embed it in the code.

autosportlabs/racecapture/views/util/alertview.py
brentpicasso: it's a helper container for a simple editor dialog.

autosportlabs/racecapture/views/util/viewutils.py
brentpicasso: rabble!  fixed.

autosportlabs/uix/track/racetrackview.py
brentpicasso: fixed!

autosportlabs/uix/track/trackmap.py
brentpicasso: fixed throughout file (even old stuff)

autosportlabs/help/helpmanager.py
brentpicasso: I'm keen on optimization. I'm wondering if they will bulk up the in memory executable (if the static strings linger around.. alternatively, since they're in memory already, it could load faster than if they loaded from a file.  We should profile it!

autosportlabs/racecapture/datastore/datastore.py
brentpicasso: \o/  feedback to user!

autosportlabs/racecapture/settings/prefs.py
brentpicasso: fixed the one added. may do the other ones

autosportlabs/help/helpmanager.py
ryandoherty: Yeah I'm fine with this direction, makes it a lot easier to work on UI components.

autosportlabs/help/helpmanager.py
ryandoherty: Oh, I'm just talking about changing the `help_popup.foo = bar` assignments to `HelpBubble(foo=bar, baz)`, etc.

autosportlabs/racecapture/datastore/datastore.py
ryandoherty: IntelliJ can do refactorings to handle this ;)

autosportlabs/racecapture/datastore/datastore.py
ryandoherty: I don't think we should have multiple databases to store different groups of sessions. That would add a lot of complexity to the codebase and UI to manage it. We have a database, we can use labels, groups or some type of Many:Many way of managing groups of sessions. I've used systems that have identical tables and databases to group things, ends up being overly complex.

autosportlabs/racecapture/datastore/datastore.py
ryandoherty: Is it used in this method? I can't find it.

autosportlabs/racecapture/datastore/datastore.py
ryandoherty: So, what does it do? Does it toss the data? I'm wondering how would the user know what happened and how to fix the problem (if possible)

autosportlabs/racecapture/datastore/datastore.py
ryandoherty: :+1:

autosportlabs/racecapture/datastore/datastore.py
ryandoherty: You just haven't used a good one yet :) Plus the good ones have support for more complex operations. Our schema and use is fairly simplistic (min/max/avg, search, few joins). 

Either direction you take will have gotchas. String building is brittle and doesn't allow for easy modification or reuse. ORMs can add bloat and sometimes can't handle complex queries. 

I don't think we need one now, just thinking. SQLAlchemy is neat, it has a separate SQL query builder and ORM, so you can use either. http://www.sqlalchemy.org/ Used by Reddit, Mozilla and Yelp. 

At minimum if we start having problems or code duplication because we're building strings we should take a look at other options.

autosportlabs/racecapture/views/analysis/analysisview.py
ryandoherty: :+1:

autosportlabs/racecapture/views/analysis/sessionbrowser.py
ryandoherty: Sounds good.

autosportlabs/racecapture/views/analysis/sessionbrowser.py
ryandoherty: The alert popup is fine, catching all Exceptions is what I'm talking about. Catching all exceptions can mean you'll catch things you weren't expecting, which could cause weird behavior, especially if you attempt any actions in the exception handler.

autosportlabs/racecapture/views/util/alertview.py
ryandoherty: Oops, I meant what is `__all__` . I googled it, :+1:

resource/help/help_text.json
ryandoherty: For localization we should use gettext (https://docs.python.org/2/library/gettext.html), which is the defacto standard. We could even start now I think, because gettext will just return the English string if there are no localization files available.

autosportlabs/help/helpmanager.py
brentpicasso: Ok, updated. passing more things via constructor. good idea.

autosportlabs/racecapture/datastore/datastore.py
brentpicasso: Refactored! changed ses_id to session_id

autosportlabs/racecapture/datastore/datastore.py
brentpicasso: I'll go with that. Agree, it reduces the need have a 'database selection' interface and all of the associated code.  we can add labels and groupings later as people want to manage sessions from different tracks.

autosportlabs/racecapture/datastore/datastore.py
brentpicasso: We're not yet presenting them to the user. on the short term ToDo (may add it to this PR or in 1.4.1)

autosportlabs/racecapture/datastore/datastore.py
brentpicasso: It skips over that line and doesn't import it. The best we can do is tell the user what lines in the import had invalid data.

autosportlabs/racecapture/datastore/datastore.py
brentpicasso: fixed.

autosportlabs/racecapture/datastore/datastore.py
brentpicasso: fixed

autosportlabs/racecapture/datastore/datastore.py
brentpicasso: The datastore uses generic exceptions throughout; I will replace with a specific DatastoreException.

brentpicasso: Fixed.

autosportlabs/racecapture/datastore/datastore.py
ryandoherty: :+1:

autosportlabs/racecapture/datastore/datastore.py
brentpicasso: fixed.

autosportlabs/racecapture/datastore/datastore.py
ryandoherty: Ok. I guess long-term some way of reporting bad data to the user will be necessary.

autosportlabs/racecapture/views/analysis/addstreamview.kv
brentpicasso: FINE :D

brentpicasso: updated.

ryandoherty: \o/

autosportlabs/racecapture/views/analysis/addstreamview.py
brentpicasso: fixed.

autosportlabs/racecapture/views/analysis/channelvaluesview.py
brentpicasso: ?

autosportlabs/racecapture/views/analysis/sessionbrowser.py
brentpicasso: switched to use new DatastoreException

autosportlabs/racecapture/views/analysis/channelvaluesview.py
ryandoherty: `self.channel_stats = {}` :D

brentpicasso: Gotcha. Also, we need to run this all through a python file formatter.
Would be good cleanup!

On Fri, Jan 8, 2016 at 1:05 PM, Ryan Doherty <notifications@github.com>
wrote:

> In autosportlabs/racecapture/views/analysis/channelvaluesview.py
> <https://github.com/autosportlabs/RaceCapture_App/pull/778#discussion_r49238049>
> :
>
> > +
> > +    @property
> > +    def maxval(self):
> > +        return self.value_view.maxval
> > +
> > +    @maxval.setter
> > +    def maxval(self, value):
> > +        self.value_view.maxval = value
> > +
> > +
> > +class ChannelValuesView(ChannelAnalysisWidget):
> > +    color_sequence = ObjectProperty(None)
> > +
> > +    def __init__(self, **kwargs):
> > +        super(ChannelValuesView, self).__init__(**kwargs)
> > +        self.channel_stats={}
>
> self.channel_stats = {} :D
>
> —
> <https://github.com/autosportlabs/RaceCapture_App/pull/778/files#r49238049>
> .
>



-- 
Brent Picasso

autosportlabs/racecapture/views/analysis/addstreamview.py
brentpicasso: the file_connect / cloud_connect / wireless_connect are fired by a sub-view handled by the screen manager, which causes the parent to select what the next thing is. 

In the case of 'connect_stream_complete' this parent view forwards the event up to the analysis view.

In retrospect, I think we could have a single event with a param indicating what type of stream wants to be connected; would reduce number of event types and handler functions.

brentpicasso: Ok, simplified the eventing around selecting different stream inports

autosportlabs/racecapture/views/analysis/addstreamview.py
brentpicasso: simplified (see above)

autosportlabs/racecapture/views/analysis/addstreamview.py
brentpicasso: Went one beyond; eliminated popup. if user erases the session name, replace it again with the default. 

Also, the import log button is disabled until conditions are valid for importing.

autosportlabs/racecapture/views/analysis/addstreamview.py
brentpicasso: Fixed per comment above.

autosportlabs/racecapture/views/dashboard/comboview.py
brentpicasso: Fixed, and more. (fixed other dashboard screens as well, including PEP8 compliant fixes)

autosportlabs/racecapture/views/dashboard/comboview.py
brentpicasso: fixed. left over crap

autosportlabs/racecapture/views/dashboard/widgets/imugauge.py
brentpicasso: Fixed. Refactoring touched a lot of classes in the dashboard view, though.

autosportlabs/racecapture/views/channels/channelselectview.py
brentpicasso: Sure. It's mostly around making it work with kivy's listadapter, which is in Kivy docs, but added some docs to explain the kivy voodoo.

brentpicasso:
Regarding Datastore's DDL, I think we should keep it in the same file for reasons similar to the .kv file. 

What I can do is define it at the top in an array so it's easier to read, and then the _create_tables() call can loop through that array.

brentpicasso:
Per conversation, we'll handle the schema management via separate PR.

resource/help/help_text.json
brentpicasso: I think we can still use a centralized collection of help messages to make it easier to manage (this would be expand to many of the features of the app; popup messages, interactive guides, like we're starting with in the analysis features).

It can still be made gettext compatible later and still use this mechanism.

autosportlabs/racecapture/datastore/datastore.py
ryandoherty: I still think we need at least a query builder long-term. You can do it this way, but it's not easy and ends up being harder in the long run when you start needing to change queries after being built.

brentpicasso: agreed. let's cross this bridge when it's pragmatic.  The datastore interface itself has a nice query builder, focused on the lap data, we can continue that pattern at the database layer as well.

ryandoherty:
Overall this looks good! Very Pythonic.

Outstanding issues are:

- Still very few to no comments for all code. I'm not asking for lots of dumb comments, just if there was anything particularly complex or crazy that had to be done, a comment about it could help the next dev who may have to update things. Or comments that describe things that aren't obvious.
- High-level comments/descriptions of how to use the visualization classes (map, chart), HelpBubble, or other re-usable/extendable classes that will be reused later on. A short 3 or 4 line code sample of how to init, add data and render would be great. Otherwise incoming devs (myself included) will have to do lots of code spelunking to figure out how it works.
- Renaming unit tests to run in a specific order is usually a sign of weird dependencies. Unit tests should work in any order.

brentpicasso:
Good feedback, will take care of this on next set of commits.

On Sat, Jan 9, 2016 at 5:41 PM, Ryan Doherty <notifications@github.com>
wrote:

> Overall this looks good! Very Pythonic.
>
> Outstanding issues are:
>
>    - Still very few to no comments for all code. I'm not asking for lots
>    of dumb comments, just if there was anything particularly complex or crazy
>    that had to be done, a comment about it could help the next dev who may
>    have to update things. Or comments that describe things that aren't obvious.
>    - High-level comments/descriptions of how to use the visualization
>    classes (map, chart), HelpBubble, or other re-usable/extendable classes
>    that will be reused later on. A short 3 or 4 line code sample of how to
>    init, add data and render would be great. Otherwise incoming devs (myself
>    included) will have to do lots of code spelunking to figure out how it
>    works.
>    - Renaming unit tests to run in a specific order is usually a sign of
>    weird dependencies. Unit tests should work in any order.
>
> —
> <https://github.com/autosportlabs/RaceCapture_App/pull/778#issuecomment-170299775>
> .
>



-- 
Brent Picasso

PR #788, opened 2016-01-08
Use working ihexpy module

Knio:


ryandoherty:
Nice, thanks for catching this :+1:

PR #793, opened 2016-01-10
Win10 firmware update fix

brentpicasso:
Addresses the issue with firmware programming on windows 10 (and possibly windows 8).  Main source of the fix was to update to pyserial 3.0, which seems to address certain windows 10 issues.

Also cleaned up logic around disabling API while firmware updating; improved logging. 

Also generated new asl_f4_loader so logger can be passed to FwUpdater and messages can be appear in kivy log vs just via print statements.


autosportlabs/racecapture/views/configuration/rcp/firmwareupdateview.py
```diff
@@ -164,8 +166,7 @@ def _update_thread(self, instance):
         except Exception as detail:
             alertPopup('Error Loading', 'Failed to Load Firmware:\n\n' + str(detail))
 
-        if not platform == 'win':
-            self._restart_json_serial()
```
brentpicasso: we want to restart json API on all platforms, don't know why this was here originally...

brentpicasso:
Related PR for ASL_F4_loader: https://github.com/autosportlabs/ASL_F4_bootloader/pull/4

autosportlabs/comms/serial/serialconnection.py
```diff
@@ -26,7 +26,7 @@ def isOpen(self):
         return self.ser != None
     
     def open(self, port):
-        ser = serial.Serial(port, timeout=self.timeout, writeTimeout = self.writeTimeout) 
+        ser = serial.Serial(port, timeout=self.timeout, write_timeout = self.writeTimeout)
```
brentpicasso: so... pyserial changed their API. made it PEP8 compliant, apparently.

ryandoherty: So, were were including/requiring pyserial without specifying the version? I'm confused why we have to change this for version 3.0.

ryandoherty: Oops, nvm, I see the version update.

brentpicasso: when I bumped pyserial from 2.7 to 3.0 this line broke, because they changed the interface.

ryandoherty:
Looks good to me :+1:

PR #804, opened 2016-01-12
Support any RaceCapture bluetooth device name

brentpicasso:
Allow the android device to pair with multiple RaceCapture bluetooth devices, each with a unique name. The app will connect with the first device it can, regardless of the name. 

Currently the app looks for a fixed RaceCapturePro name. this change will make it compatible with 2.9.0 firmware changes.

### DIFF
--- a/autosportlabs/comms/androidcomms.py
+++ b/autosportlabs/comms/androidcomms.py
@@ -32,24 +32,26 @@ class AndroidComms(object):
         self._bt_conn = BTConn.createInstance();
                                                                 
     def get_available_ports(self):
-        return ['RaceCapturePro'] #TODO get this from the service directly
+        bt_ports = self._bt_conn.getAvailableDevices()
+        Logger.info('AndroidComms: detected ports: {}'.format(bt_ports.join(',')))
+        return bt_ports
     
     def isOpen(self):
         return self._bt_conn.isOpen()
     
     def open(self):
-        Logger.info('androidcomms: Opening connection ' + str(self.port))
+        Logger.info('AndroidComms: Opening connection ' + str(self.port))
         self._bt_conn.open(self.port)
-        Logger.info('androidcomms: after open!!!!!!!!!!!!!!!!')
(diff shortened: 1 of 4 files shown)


brentpicasso:
This PR addresses #804

autosportlabs/comms/androidcomms.py
```diff
@@ -32,24 +32,26 @@ def __init__(self, **kwargs):
         self._bt_conn = BTConn.createInstance();
                                                                 
     def get_available_ports(self):
-        return ['RaceCapturePro'] #TODO get this from the service directly
```
ryandoherty: Any benefit for this to be an `@property`?

autosportlabs/comms/androidcomms.py
```diff
@@ -32,24 +32,26 @@ def __init__(self, **kwargs):
...
     
     def isOpen(self):
         return self._bt_conn.isOpen()
     
     def open(self):
-        Logger.info('androidcomms: Opening connection ' + str(self.port))
+        Logger.info('AndroidComms: Opening connection ' + str(self.port))
         self._bt_conn.open(self.port)
-        Logger.info('androidcomms: after open!!!!!!!!!!!!!!!!')
+        Logger.info('AndroidComms: after open!!!!!!!!!!!!!!!!')
```
ryandoherty: Think we could remove 15 or 16 exclamation points while we're here? :P

brentpicasso: lol that is so old. I was excited when it worked :)

ryandoherty:
Overall looks fine to commit with our without suggested changes.

autosportlabs/comms/androidcomms.py
brentpicasso: IDK. since it's digging into library code each time for the answer it's doing more than encapsulating a single variable for quick access. Even so, it's idempotent and should be fine as a property. Making the change would require changes across serial api and across other classes, and I am le tired.

I guess the function could be called 'query_available_ports'....

PR #820, opened 2016-01-13
P/brentpicasso/async query import perf bonanza

brentpicasso:
* Refactor / improve import functions
* Improve channel data querying performance (removed redundant table-scan queries)
* Async querying / loading to prevent blocking of UI. show spinner as needed


autosportlabs/racecapture/datastore/datastore.py
```diff
@@ -646,10 +648,16 @@ def _handle_data(self, data_file, headers, session_id, warnings=None, progress_c
...
-            self._insert_record(record, headers, session_id, warnings=warnings)
-
-        self._conn.commit()
+        datalog_id = self._get_last_table_id('sample') + 1
+        
+        cur = self._conn.cursor()
+        try:
+            for record in newdata_gen:
+                datalog_id = self._insert_record(cur, datalog_id, record, headers, session_id, warnings=warnings)
+            self._conn.commit()
```
ryandoherty: Did this speed up anything? Got benchmarks?

autosportlabs/racecapture/datastore/datastore.py
```diff
@@ -646,10 +648,16 @@ def _handle_data(self, data_file, headers, session_id, warnings=None, progress_c
...
+        except:
```
ryandoherty: Generic exception handler! :)

autosportlabs/racecapture/views/analysis/analysisdata.py
```diff
@@ -9,27 +12,48 @@ def __init__(self, **kwargs):
...
     
     def __init__(self, **kwargs):
-        self.data = kwargs.get('data', None)
+        self.values = kwargs.get('values', None)
         self.channel = kwargs.get('channel', None)
         self.min = kwargs.get('min', 0)
         self.max = kwargs.get('max', 0)
         self.source = kwargs.get('source', None)
         
+class ChannelDataParams(object):
```
ryandoherty: I'm kinda confused as to what this is for. I see it's used for querying data? Maybe some comments here about how to use it and why.

autosportlabs/racecapture/views/analysis/analysisview.py
```diff
@@ -81,7 +80,7 @@ def on_channel_selected(self, instance, value):
 
     def on_marker(self, instance, marker):
         source = marker.sourceref
-        cache = self._datastore.get_location_data(source)
+        cache = self._datastore.get_cached_location_data(source)
```
ryandoherty: I'm wondering if the consumers of the Datastore need to know whether or not they are fetching cached data. Does it matter? Why not let the underlying class figure out to get it from the cache or not?

ryandoherty:
Overall structure seems fine. A few questions.

brentpicasso:
Added some actual optimizations by using a generator to speed up the sqlite inserts. 1.8x faster when importing the sonoma.log file.

with generator: [Datastore   ] import_datalog function took 3710.3638649 ms
r/1.4.2         [Timing      ] import_datalog function took 6680.72295189 ms

1.8x faster

autosportlabs/racecapture/views/analysis/analysisdata.py
brentpicasso: It's a class to hold parameters passed to the query worker. see line ~124.  Will add doc tag.

autosportlabs/racecapture/views/analysis/analysisview.py
brentpicasso: get_location_data was changed to behave in an async way, by passing the request to a thread via a queue.  

With the marker (when you mouse over the chart) it will call this function at a high frequency. We want to quickly access cached data (if available). 

 If not available, the marker won't work.  But, that's ok b/c under those conditions we're likely still loading data. It's designed to be tolerant of the cached data not being present.

autosportlabs/racecapture/datastore/datastore.py
brentpicasso: Nothing dramatic, just plain refactoring (wrapping in a transaction).  However i pushed a version that uses a generator, which made a more dramatic perf difference.

autosportlabs/racecapture/views/analysis/analysisview.py
ryandoherty: I'm still thinking that consumers of a DataStore's data shouldn't have to use 2 different functions `get_location_data` or `get_cached_location_data`, they don't care if it's cached or not, they just want the data.

A similar example is how a website may cache data. Users or software don't know if it's cached or not, they just hit a URL and get data.

ryandoherty:
:+1: on the speedup!

autosportlabs/racecapture/views/analysis/analysisview.py
brentpicasso: Agreed. I collapsed it into one function that can handle dealing with a cache miss. Thanks for the feedback!

autosportlabs/racecapture/datastore/datastore.py
```diff
@@ -637,19 +632,50 @@ def _create_session(self, name, notes=''):
         Logger.info('DataStore: Created session with ID: {}'.format(session_id))
         return session_id
 
+
+    #class member variable to track ending datalog id when importing
+    _ending_datalog_id  = 0
```
ryandoherty: This would end up being a static variable shared across all DataStore instances, is that intentional?

brentpicasso: We could make it non static. What I really wanted was to make it a local variable if possible.

ryandoherty: Why does it need to be declared outside the _handle_data function? If it's only used inside that function it can be declared in there, correct? I'm wondering if I'm not seeing something about this var.

brentpicasso: http://stackoverflow.com/questions/5218895/python-nested-functions-variable-scoping

Python3 has the nonlocal keyword, but we're on python 2.7

ryandoherty: Ah, I see, the nested function. If we wrote this in JavaScript this would work :trollface:

brentpicasso: http://rack.2.mshcdn.com/media/ZgkyMDEzLzA2LzEzL2Y0L1B1Y2suODkyZGIuZ2lmCnAJdGh1bWIJMTIwMHg5NjAwPg/c6538ddb/a38/Puck.gif

ryandoherty: So a Python expert would probably just say to declare it in the constructor. While the chance is low that our code would have 2 DataStores and be importing at the same time, it's still probably better to code it in a safe manner.

brentpicasso: agreed.

brentpicasso: I was so annoyed I couldn't make it local I really wanted to put it next to the function :)

ryandoherty: :hug:

brentpicasso:
Refactored to batch query channels, major perf improvement when browsing. Added more documentation.

Issue #762, opened 2015-12-17
when importing a log file, automatically extend min / max's for channels


brentpicasso:
If a log file is imported with nonsensical metadata (e.g. very small max value for speed) adjust the datastore's min/max values to allow data to fit and therefore show correctly in line charts and other views.

PR #828, opened 2016-01-17
Automatically extend min/max metadata upon logfile import

brentpicasso:
When importing a logfile, automatically extend the min/max metadata values as necessary. This allows charts to show up correctly if the datalog data extends beyond the range of the current min/max.


brentpicasso:
addresses #762

ryandoherty:
Seems ok, my thoughts are:

- Could these values be computed on import while we read the log file without querying all rows? Would be less work.
- What happens when a user removes a session? Do we do this update again or leave min/maxes?

brentpicasso:
we need this as a general purpose tool to adjust min/max channel values in future functionality, so I didn't want to weave it into the import process. 

If they remove a session nothing happens. The primary goal at this point is to ensure values aren't clipped after importing. In the future the extended options will be used to let the user automatically rail the min/max values to the actual values in the file, or letting them auto-reset them to system defaults, or edit the values directly.

ryandoherty:
Ok, cool. :shipit:

Issue #818, opened 2016-01-13
Instruct user to save current config before updating firmware


brentpicasso:
before updating firmware, app should pop up a message instructing users to save their current config

stieg:
Why not just have the app do it automatically as part of the upgrade process?  Perhaps just save it to a temp file to write back into RCP after the upgrade?  So the process would go:

* Download config to a temp file
* Upgrade firmware
* Upon successful upgrade, write config back.

That way it happens transparently.  Bonus points if it doesn't mess with the config UI screen.  Of course... not sure how complicated of a task that might be.

brentpicasso:
Yes, this will be covered in #817 .   The configuration infrastructure also need to be worked over to handle multiple different devices (RCP, RC, etc),  #817 can get folded in to that effort in one shot.

This is just a quick PSA in the meantime, since we have nothing at the moment.

stieg:
Got it.  Makes sense.

PR #830, opened 2016-01-18
Improvements to firmware update screen

brentpicasso:
* Prompt user to save config as necessary
* Hid not working button (check for update online)


brentpicasso:
fixes #818

ryandoherty:
The popup happens after the user has select the firmware file, I think it should show up when they click on the 'Update Firmware' button. That way they don't waste time finding their firmware image to then realize they need to backup their config. Thoughts?

brentpicasso:
Agreed. flipped the order and made other improvements. thank you.

ryandoherty:
:shipit:

Issue #835, opened 2016-01-19
Remove obsolete CurrentLapTimeFilter


brentpicasso:
It seems that the logic in currentlaptimefilter.py may be obsolete with the latest functionality in laptime.py

Investigate / verify and then remove if necessary.

PR #836, opened 2016-01-19
removed obsolete currentLaptime; fixed metadata for elapsedtime

brentpicasso:
The logic in the laptime widget now correctly chooses between what channels to show for the current laptime (elapsed time, predicted time, or last lap time) so we don't need this channel filter anymore.


brentpicasso:
addresses #835

ryandoherty:
:shipit:

Issue #832, opened 2016-01-18
Show track name selected in map view


brentpicasso:
No way to know the exact track map shown in track map view.

Issue #774, opened 2015-12-29
Show a color legend of the selected lap in the map view


brentpicasso:
In the map view, show a legend for which session / lap is selected and the selected color

e.g.

<session> <lap>  [COLOR BAR]

Issue #773, opened 2015-12-29
Show heatmap legend when in heatmap mode


brentpicasso:
When trackmap is in heatmap mode, show the selected channel, a color gradient and the min / max values

e.g.  

TPS  0 [GRADIENT COLOR BAR] 100

PR #840, opened 2016-01-22
Analysis map enhancements

brentpicasso:
Add additional information to analysis map view. Addresses #832, #773, #774

* Name of selected track map
* Legends for selected track paths
* Heatmap legends when heat map mode is enabled (including selected heatmap channel)
* Multi-colored heatmap when multiple laps are selected, to help visually differentiate heatmaps
* Updated doctags
* Added GPL license headers


autosportlabs/racecapture/datastore/datastore.py
```diff
@@ -291,7 +291,13 @@ def is_open(self):
     def channel_list(self):
         return self._channels[:]
 
-    def get_channel(self, name): 
+    def get_channel(self, name):
+        '''
+        Retreives information for a channel
```
ryandoherty: Typo.

ryandoherty:
Overall the code looks good, comments are great :+1: (Although maybe next time do the license stuff in a separate branch :smile: )

Notes: (many of these may be unrelated, we can file them separately if needed)

Major:

It's still really hard to understand the heat map colors when more than 1 lap is shown. Especially since the colors are very close to one another: http://i.imgur.com/pSgfmbE.png And when they aren't close to one another my eyes still have to jump back and forth at least a few times to understand what it all means: http://i.imgur.com/HCWvgco.png  I don't think we should ship it this way, using one gradient for each lap would make it much more understandable. Overlapping for now is ok, but parallel would be much better.

Some laps take a long time to load:

```
[INFO   ] [Datastore   ] _query_channel_data function took 13464.0059471 ms
[INFO   ] [CachingAnalysisDatastore] querying 15_1 ['Distance', 'Speed']
[INFO   ] [Datastore   ] _query_channel_data function took 19.6471214294 ms
[INFO   ] [CachingAnalysisDatastore] querying 14_1 ['Distance', 'Speed']
[INFO   ] [Datastore   ] _query_channel_data function took 28836.1179829 ms
```
I think this is due to data from before/after a lap completes being loaded. Using a combination of LapCount and CurrentLap you can determine most laps true start/finish as opposed to just using LapCount.

Performance starts to lag when you have the heatmap on and 3 or more laps loaded.

Minor:
Scroll events over the list of channels to select when configuring the heat map affect the map. You have to use the scroll wheel over top part of the list that overlays on top of the map.

Graph doesn't resize when you unload a lap that had a lot longer distance than the rest: http://i.imgur.com/Ln66c1f.png

Deleting the session while laps are loaded results in nothing being removed: http://i.imgur.com/MUjm21K.png (I think already filed)

ryandoherty:
Testing on a LG-V500, heatmap really slows down the app, very laggy. Very slow on a Nexus 6P too.

brentpicasso:
changed heat map to same color range for all selected laps; some other refactorings.

ryandoherty:
The only thing left that seems related and is a bug is:

Scroll events over the list of channels to select when configuring the heat map affect the map. You have to use the scroll wheel over top part of the list that overlays on top of the map.

brentpicasso:
Thanks. Was going to handle that separately but can certainly bundle it in here! :)

ryandoherty:
Doesn't need to be bundled, wasn't sure. I can file all the other issues separately.

brentpicasso:
Feel free to create the issue, I'll add it to this PR . I thought there was
an issue for this already, but couldn't find it via casual search.

On Fri, Jan 29, 2016 at 9:31 AM, Ryan Doherty <notifications@github.com>
wrote:

> Doesn't need to be bundled, wasn't sure. I can file all the other issues
> separately.
>
> —
> <https://github.com/autosportlabs/RaceCapture_App/pull/840#issuecomment-176873759>
> .
>



-- 
Brent Picasso

PR #847, opened 2016-01-29
Add logic that prevents StatusPump from having more than one instance

stieg:
Because more than once instance creates unnecessary traffic and threads.  And that makes MK1 a sad panda.

### DIFF
--- a/autosportlabs/racecapture/status/statuspump.py
+++ b/autosportlabs/racecapture/status/statuspump.py
@@ -27,7 +27,8 @@ class StatusPump(object):
         self._listeners.append(listener)
 
     def start(self, rc_api):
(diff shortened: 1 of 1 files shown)


autosportlabs/racecapture/status/statuspump.py
```diff
@@ -10,31 +10,35 @@
...
     _status_thread = None
-    
+
     def add_listener(self, listener):
         self._listeners.append(listener)
-                    
+
     def start(self, rc_api):
-        Logger.info('StatusPump: starting')
+        if self._status_thread != None and self._status_thread.is_alive():
```
brentpicasso: Nit: a bit more pythonic to say is not None

brentpicasso:
one minor nit - otherwise :shipit:

PR #848, opened 2016-01-29
Only show active channels in the selection list when RCP is connected

### LINKED ISSUE #829: Channel selection should only show active sensor channels
When selecting a channel for display in the dashboard, should only show channels actually available in the data stream. Additionally, system channels like Time, GPS sensors, etc should not be in the list.

Currently all channels are displayed for selection and it creates un-necessary clutter.

brentpicasso:
Only show active channels in the selection list when RCP is connected - otherwise show the full list of system channels. resolves #829

### DIFF
--- a/autosportlabs/racecapture/views/channels/channelselectview.py
+++ b/autosportlabs/racecapture/views/channels/channelselectview.py
@@ -33,12 +33,12 @@ class ChannelSelectorView(BoxLayout):
         if channels:
             index = 0
             list_adapter = self.ids.channelList.adapter
-            for item in self.ids.channelList.adapter.data:
+            for item in list_adapter.data:
                 if item['text'] in channels:
                     view = list_adapter.get_view(index)
                     view.trigger_action(duration=0) #duration=0 means make it an instant selection
                 index += 1
-        
+
     def on_channels(self, instance, value):    
         '''
         Set the list of available channels for this view.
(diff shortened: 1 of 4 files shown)


autosportlabs/racecapture/views/channels/channelselectview.py
```diff
@@ -95,7 +92,16 @@ def __init__(self, **kwargs):
                                selection_mode='single',
                                allow_empty_selection=True)
     
-            channel_list.adapter=list_adapter
+            channel_list.adapter = list_adapter
+
+            #select the current channel
+            index = 0
+            for item in list_adapter.data:
```
ryandoherty: From the Kivy docs, ListAdapter has a `selection` property, could that be used instead of iterating the values?

ryandoherty:
Overall looks really good, just one question.

autosportlabs/racecapture/views/channels/channelselectview.py
brentpicasso: No... you'd think should work that way, but the voodoo in the code is how we're supposed to do it.

ryandoherty: Wat. Then what's the benefit of using a ListAdapter?

brentpicasso: It's just the container for holding the model. <shrug>

brentpicasso: http://stackoverflow.com/questions/19459808/kivy-select-an-item-in-listview-from-code

ryandoherty: Well that's silly. Ok.

ryandoherty:
Very nice, :shipit:

PR #849, opened 2016-01-29
Draw path if no trackmap

### LINKED ISSUE #834: Show GPS data if no trackmap is provided
If no track map is detected, still show just the GPS trace in the analysis view.

brentpicasso:
If a trackmap is not loaded, still show the GPS path in the trackmap view.

This change causes the path to be scaled correctly when loaded into the trackmap view.

resolves #834


ryandoherty:
:shipit:

PR #857, opened 2016-01-29
P/stieg/tx timeout ++

stieg:
Increases the Tx timeout so that MK1 can have the time it needs to ingest all the data.  Also adds exception chaining support to some of our exceptions so we can see what caused an underlying exception (as we wrap some exceptions for odd reasons).

When this app gets moved to python3 we can use the `from e` method on a `raise` invocation and remove the chaining code I hacked in.

### DIFF
--- a/autosportlabs/comms/commscommon.py
+++ b/autosportlabs/comms/commscommon.py
@@ -3,7 +3,7 @@ class ChainableException(Exception):
         m = ''
 
         if cause is not None:
-            m += 'caused by %s' % repr(cause)
+            m += 'caused by {}'.format(repr(cause))
             m.strip()
 
         super(Exception, self).__init__(m)
--- a/autosportlabs/comms/commscommon.py
+++ b/autosportlabs/comms/commscommon.py
@@ -1,5 +1,20 @@
-class PortNotOpenException(Exception):
+class ChainableException(Exception):
+    def __init__(self, cause=None):
+        m = ''
+
+        if cause is not None:
+            m += 'caused by %s' % repr(cause)
+            m.strip()
+
+        super(Exception, self).__init__(m)
+        self.cause = cause
+
+    def get_cause(self):
+        return self.cause
+
+
+class PortNotOpenException(ChainableException):
     pass
 
-class CommsErrorException(Exception):
+class CommsErrorException(ChainableException):
     pass
--- a/autosportlabs/comms/serial/serialconnection.py
+++ b/autosportlabs/comms/serial/serialconnection.py
@@ -26,7 +26,7 @@ class SerialConnection():
         return self.ser != None
 
     def open(self, port):
-        self.ser = serial.Serial(port, timeout=self.timeout, \
+        self.ser = serial.Serial(port, timeout=self.timeout,
                             write_timeout = self.writeTimeout)
 
     def close(self):
--- a/autosportlabs/comms/serial/serialconnection.py
+++ b/autosportlabs/comms/serial/serialconnection.py
@@ -26,9 +26,8 @@ class SerialConnection():
         return self.ser != None
 
     def open(self, port):
-        ser = serial.Serial(port, baudrate=115200, timeout=self.timeout, \
+        self.ser = serial.Serial(port, timeout=self.timeout, \
                             write_timeout = self.writeTimeout)
-        self.ser = ser
 
     def close(self):
         if self.ser != None:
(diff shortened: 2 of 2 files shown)


autosportlabs/comms/commscommon.py
```diff
@@ -1,5 +1,20 @@
-class PortNotOpenException(Exception):
+class ChainableException(Exception):
+    def __init__(self, cause=None):
+        m = ''
+
+        if cause is not None:
+            m += 'caused by %s' % repr(cause)
```
brentpicasso: use "... {} ...".format(my_var) instead of % please.

autosportlabs/comms/serial/serialconnection.py
```diff
@@ -5,30 +5,31 @@
...
         return filtered_ports
-            
+
     def isOpen(self):
         return self.ser != None
-    
+
     def open(self, port):
-        ser = serial.Serial(port, timeout=self.timeout, write_timeout = self.writeTimeout)
+        ser = serial.Serial(port, baudrate=115200, timeout=self.timeout, \
```
brentpicasso: remove baudrate please. doesn't apply to USB

stieg: Oops... thanks

autosportlabs/comms/commscommon.py
stieg: TIL about `.format`.  Will do.

brentpicasso:
looks legit. How does MK2 like it?

Also, would need to test on android - but changes to commscommon shouldn't have impact. I can do the test.

autosportlabs/comms/serial/serialconnection.py
```diff
@@ -5,30 +5,30 @@
...
+
     def isOpen(self):
         return self.ser != None
-    
+
     def open(self, port):
-        ser = serial.Serial(port, timeout=self.timeout, write_timeout = self.writeTimeout)
-        self.ser = ser
-            
+        self.ser = serial.Serial(port, timeout=self.timeout, \
```
brentpicasso: I think you can remove the slash or put it all on one line?

stieg: I guess I can remove the '\'.  I can never remember when you need to use a slash in python.

Wrapped it because otherwise it violates PEP8.  See https://www.python.org/dev/peps/pep-0008/#maximum-line-length

stieg:
MK2 is quite happy with it.

autosportlabs/comms/serial/serialconnection.py
brentpicasso: Okay. Although "this team" prefers the alternative longer line lengths (100) putting it on one line would still exceed that.  thanks.

brentpicasso:
will test with android and then if that works will merge. thank you!

autosportlabs/comms/serial/serialconnection.py
stieg: According to our own CONTRIBUTING.md file:

```
We use PEP8 for our coding standard. Note, much of our older code is non PEP8 compliant; the goal is to update to PEP8 as we touch those areas.
```

That is what I followed :D

brentpicasso: The PEP8 docs say that 100 is an acceptable alternative - that's what I was
referring to. :)

On Fri, Jan 29, 2016 at 2:52 PM, Andrew Stiegmann <notifications@github.com>
wrote:

> In autosportlabs/comms/serial/serialconnection.py
> <https://github.com/autosportlabs/RaceCapture_App/pull/857#discussion_r51327133>
> :
>
> >      def open(self, port):
> > -        ser = serial.Serial(port, timeout=self.timeout, write_timeout = self.writeTimeout)
> > -        self.ser = ser
> > -
> > +        self.ser = serial.Serial(port, timeout=self.timeout, \
>
> According to our own CONTRIBUTING.md file:
>
> We use PEP8 for our coding standard. Note, much of our older code is non PEP8 compliant; the goal is to update to PEP8 as we touch those areas.
>
> That is what I followed :D
>
> —
> <https://github.com/autosportlabs/RaceCapture_App/pull/857/files#r51327133>
> .
>



-- 
Brent Picasso

stieg: I assume you are referring to this:

> Some teams strongly prefer a longer line length. For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the nominal line length from 80 to 100 characters (effectively increasing the maximum length to 99 characters), provided that comments and docstrings are still wrapped at 72 characters. 

I prefer 80 but I'm not about to start a coding religious war here.  100 is fine if you stick to the qualifications that PEP8 sets.  Please update the CONTRIBUTING doc accordingly so it represents the choice

brentpicasso:
works fine on android! merging.

brentpicasso:
aw crap, this is still against master...

PR #858, opened 2016-01-29
Add print Tx/Rx serial stream arg to app without setting Trace

stieg:
More than a few times I have needed a way to see what was being sent
to and from RCP.  Before this patch the only way I could do this was
to enable Trace level logging (which is not easy to do unfortunately).
But the problem with this is that if I enable Trace level debugging, I
get TONS of logs.  Metric TONS.

That is not helpful to me.  So instead I wrote this patch that will
allow me to extract out the Tx/Rx data by passing the following arg

./main.py -- --print_tx_rx

### DIFF
--- a/autosportlabs/racecapture/api/rcpapi.py
+++ b/autosportlabs/racecapture/api/rcpapi.py
@@ -39,18 +39,18 @@ class RcpCmd:
         self.payload = payload
         self.index = index
         self.option = option
-            
+
 class CommandSequence():
     command_list = None
     rootName = None
     winCallback = None
     failCallback = None
-            
+
 class RcpApi:
     detect_win_callback = None
     detect_fail_callback = None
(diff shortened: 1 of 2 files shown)


brentpicasso:
Can we see if we can change it from trace to debug level provides a reasonable balance?

stieg:
We could.  Kivy supports passing the -d flag on the initial parameters so that satisfies my requirement of having this be trigger able from the command line.  I'll do this in another PR.

stieg:
Debug Review: autosportlabs/RaceCapture_App/pull/859

stieg:
So that works.  Only drawback is that its showing data not necessarily relevant to debugging the app itself, but that is a gray line.  If you like autosportlabs/RaceCapture_App#859 better... go with it and close this one.  Doesn't matter to me.

PR #859, opened 2016-01-29
Move Serial Tx/Rx log levels to Debug instead of Trace

stieg:
While kivy does support "Trace" level, it is not eaisly activated.
Rather one must enable in in their special config file to get it
to work.  This change moves the Tx/Rx commands to the debug level
so that we can see all the chatter between RCP and the App on the
command line.  This is enableable by passing in the '-d' argument
from the command line so its more useful.


brentpicasso:
looks good, but can it be merged into r/1.4.2? thanks.

stieg:
You want me to branch it off of 1.4.2?  Ok.

PR #861, opened 2016-01-30
use locks to prevent dictionary contention (IOW a different thread pu…

### LINKED ISSUE #852: Sentry: RuntimeError: dictionary changed size during iteration
Appears channel meta data updates are pulling the rug out from underneath notify listeners routine.

-----------------------

dictionary changed size during iteration
home/brent/git-projects/RaceCapture_App/.buildozer/android/platform/python-for-android/build/python-install/lib/python2.7/site-packages/kivy/core/window/window_pygame.py in mainloop at line 398

home/brent/git-projects/RaceCapture_App/.buildozer/android/platform/python-for-android/build/python-install/lib/python2.7/site-packages/kivy/core/window/window_pygame.py in _mainloop at line 285

home/brent/git-projects/RaceCapture_App/.buildozer/android/platform/python-for-android/build/python-install/lib/python2.7/site-packages/kivy/base.py in idle at line 327

home/brent/git-projects/RaceCapture_App/.buildozer/android/platform/python-for-android/build/python-install/lib/python2.7/site-packages/kivy/clock.py in tick at line 483

home/brent/git-projects/RaceCapture_App/.buildozer/android/platform/python-for-android/build/python-install/lib/python2.7/site-packages/kivy/clock.py in _process_events at line 615

home/brent/git-projects/RaceCapture_App/.buildozer/android/platform/python-for-android/build/python-install/lib/python2.7/site-packages/kivy/clock.py in tick at line 374

home/brent/git-projects/RaceCapture_App/.buildozer/android/app/autosportlabs/racecapture/databus/databus.py in notify_listeners at line 102

brentpicasso:
Sentry reported periodic exceptions where the channel_data dict was getting stepped on by a different thread. This puts a Lock (mutex) around accessing that dict to prevent errors from happening. 

Also, resets channel_data dict to {} if new metadata was received, in case channels were removed from the stream (due to an RCP config change).

Resolves #852 

(has other code formatting changes due to an automatic code formatting tools in the editor, sorry for the noise)

### DIFF
--- a/autosportlabs/racecapture/databus/databus.py
+++ b/autosportlabs/racecapture/databus/databus.py
@@ -8,6 +8,17 @@ from autosportlabs.racecapture.databus.filter.bestlapfilter import BestLapFilter
 from autosportlabs.racecapture.databus.filter.laptimedeltafilter import LaptimeDeltaFilter
 from autosportlabs.util.threadutil import safe_thread_exit
 
+class ThreadSafeDict(dict) :
+    def __init__(self, * p_arg, ** n_arg) :
+        dict.__init__(self, * p_arg, ** n_arg)
+        self._lock = Lock()
+
+    def __enter__(self) :
+        self._lock.acquire()
+        return self
+
+    def __exit__(self, type, value, traceback) :
+        self._lock.release()
 
 DEFAULT_DATABUS_UPDATE_INTERVAL = 0.02  # 50Hz UI update rate
 
@@ -31,7 +42,7 @@ class DataBus(object):
     Note: DataBus must be started via start_update before any data flows
     """
     channel_metas = {}
-    channel_data = {}
+    channel_data = ThreadSafeDict()
     sample = None
     channel_listeners = {}
     meta_listeners = []
@@ -40,7 +51,6 @@ class DataBus(object):
     sample_listeners = []
     _polling = False
     rcp_meta_read = False
-    channel_data_lock = Lock()
 
     def __init__(self, **kwargs):
         super(DataBus, self).__init__(**kwargs)
@@ -66,7 +76,7 @@ class DataBus(object):
         This should be called when the channel information has changed
         """
         # update
-        with self.channel_data_lock:
+        with self.channel_data as cd:
             self.channel_metas.clear()
             for meta in metas.channel_metas:
                 self.channel_metas[meta.name] = meta
@@ -76,7 +86,7 @@ class DataBus(object):
                 self._update_datafilter_meta(f)
             # clear our list of channel data values, in case channels
             # were removed on this metadata update
-            self.channel_data.clear()
+            cd.clear()
 
         self.meta_updated = True
         self.rcp_meta_read = True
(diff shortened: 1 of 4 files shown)


stieg:
This solution works but only solves the immediate problem.  There will be scenarios later where you will have thread contention for a dictionary again and need to re-implement a lock.  I suggest you implement a `ThreadSafeDict` as outlined [here](http://stackoverflow.com/questions/1312331/using-a-global-dictionary-with-threads-in-python).  Use of the `with` construct will allow you to seamlessly grab and release the lock sanely while keeping the lock where it should be, within the dictionary object.  Your code looks good, I would just move the lock to where it belongs.

brentpicasso:
Applied thread safe access to channel metas and sample data dicts throughout.  Found case where a thread in telemetryconnection was accessing the channel_meta / sample data dicts and wrapped access to dict to prevent dict from changing while a telemetry packet was being created.

brentpicasso:
recommend ?w=0  for eyeball sanity as files were auto-formatted.

autosportlabs/racecapture/databus/databus.py
```diff
@@ -57,21 +56,29 @@ def stop_update(self):
 
     def _update_datafilter_meta(self, datafilter):
         metas = datafilter.get_channel_meta()
-        for channel, meta in metas.iteritems():
-            self.channel_metas[channel] = meta
+        with self.channel_metas as cm:
```
stieg: Just curious... did you need the `as cm` argument or could you do it without it.  Either way is fine... just wanted to know what your test results were.

autosportlabs/racecapture/databus/databus.py
```diff
@@ -57,21 +56,29 @@ def stop_update(self):
...
-        self.channel_metas.clear()
-        for meta in metas.channel_metas:
-            self.channel_metas[meta.name] = meta
-            
-        #add channel meta for existing filters
-        for f in self.data_filters:
-            self._update_datafilter_meta(f)
-                
+        # update channel metadata
+        with self.channel_metas as cm:
```
stieg: Can you lock both `self.channel_metas` and `self.channel_data` in the same transaction?  That way there is no chance for threads to get weird data (albeit its already low when coded like this).

autosportlabs/util/threadutil.py
```diff
@@ -7,5 +8,17 @@
 
 def safe_thread_exit():
     if platform == 'android':
-        jnius.detach() #detach the current thread from pyjnius, else hard crash occurs
-        
\ No newline at end of file
+        jnius.detach()  # detach the current thread from pyjnius, else hard crash occurs
+
+
+class ThreadSafeDict(dict) :
```
stieg: Doesn't PEP8 say that classes like these should be in their own module (file)?

stieg:
Minor nits, but looks good.

autosportlabs/util/threadutil.py
brentpicasso: I can't find much info, only this discussion and classes don't have to be in their own files. http://programmers.stackexchange.com/questions/209982/is-it-considered-pythonic-to-have-multiple-classes-defined-in-the-same-file

autosportlabs/racecapture/databus/databus.py
brentpicasso: yeah you can totally do that! updating.

autosportlabs/racecapture/databus/databus.py
brentpicasso: not needed I think. but including it b/c seems cleaner.

autosportlabs/racecapture/databus/databus.py
stieg: You didn't update this.

autosportlabs/util/threadutil.py
stieg: Ok.

stieg:
Fix the lock around the meta and data channel bit described above and then :shipit:

autosportlabs/racecapture/databus/databus.py
brentpicasso: I didn't on purpose - the double lock thingy happens in telemetryconnection.py

stieg: Correct... but you also access both dictionaries here in this method.  I would do it here too.

brentpicasso: At first I wanted it as fine grained as possible, but in retrospect I think it would be good to lock both simultaneously due to the related updates. Thanks.

stieg: If the two dictionaries were not related then I would agree with the fine grained control.  But since they are then its wiser to control both locks. _Ideally_ both of these dictionaries would be in a larger container object under a single lock since you only modify both when the meta fields change.  Perhaps a good idea for the future.  So long as there is no deadlock here... this is fine.

brentpicasso:
Updated to have combined locks in the update_channel_meta function.

PR #865, opened 2016-01-31
Sporadic exit exception

### LINKED ISSUE #864: Sporadic exception upon application exit
An sporadic exception is occurring upon system exit, and it seems to be related to daemon threads throwing exceptions when resources are pulled out from underneath is - even if they're set to daemon.

this is showing up at a high rate in Sentry. 

This conversation seems to be the most likely explanation:
http://stackoverflow.com/questions/4079810/python-thread-exception-errors-when-exit-the-whole-program

-------------
Exception:

'NoneType' object is not subscriptable
home/brent/git-projects/RaceCapture_App/.buildozer/android/platform/python-for-android/build/python-install/lib/python2.7/site-packages/kivy/core/window/window_pygame.py in mainloop at line 398

home/brent/git-projects/RaceCapture_App/.buildozer/android/platform/python-for-android/build/python-install/lib/python2.7/site-packages/kivy/core/window/window_pygame.py in _mainloop at line 285

home/brent/git-projects/RaceCapture_App/.buildozer/android/platform/python-for-android/build/python-install/lib/python2.7/site-packages/kivy/base.py in idle at line 333

home/brent/git-projects/RaceCapture_App/.buildozer/android/platform/python-for-android/build/python-install/lib/python2.7/site-packages/kivy/lang.py in sync at line 2082

brentpicasso:
I think this resolves #864.  I was able to repro the exception upon manual exception (seemed to happen 1 out of 10 times) 

This patch has the msg_rx_worker exit when the app exits. pretty simple. After this patch I wasn't able to repro the exit exception after about 30-40 attempts.   The real tell is if the Sentry exceptions will go away or reduce dramatically.

### DIFF
--- a/autosportlabs/racecapture/api/rcpapi.py
+++ b/autosportlabs/racecapture/api/rcpapi.py
@@ -646,14 +646,14 @@ class RcpApi:
             version_result_event.set()
 
         while self._running.is_set():
+            self._auto_detect_event.wait()
+            self._auto_detect_event.clear()
+            self._enable_autodetect.wait()
+            # check again if we're shutting down
+            # to prevent a needless re-detection attempt
+            if not self._running.is_set():
+                break
             try:
-                self._auto_detect_event.wait()
-                self._auto_detect_event.clear()
-                self._enable_autodetect.wait()
-                # check again if we're shutting down
-                # to prevent a needless re-detection attempt
-                if not self._running.is_set():
-                    break
                 Logger.info("RCPAPI: Starting auto-detect")
                 self._auto_detect_busy.set()
                 self.sendCommandLock.acquire()
@@ -715,11 +715,10 @@ class RcpApi:
                 Logger.error('RCPAPI: Error running auto detect: ' + str(e))
                 Logger.debug(traceback.format_exc())
             finally:
-                if self._running.is_set():
-                    Logger.info("RCPAPI: auto detect finished. port=" + str(comms.port))
-                    self._auto_detect_busy.clear()
-                    self.removeListener("ver", on_ver_win)
-                    self.sendCommandLock.release()
-                    sleep(AUTODETECT_COOLOFF_TIME)
+                Logger.info("RCPAPI: auto detect finished. port=" + str(comms.port))
+                self._auto_detect_busy.clear()
+                self.removeListener("ver", on_ver_win)
+                self.sendCommandLock.release()
+                sleep(AUTODETECT_COOLOFF_TIME)
 
         Logger.info('RCPAPI: auto_detect_worker exiting')
--- a/autosportlabs/racecapture/api/rcpapi.py
+++ b/autosportlabs/racecapture/api/rcpapi.py
@@ -65,6 +65,8 @@ class RcpApi:
     _auto_detect_event = Event()
     _auto_detect_busy = Event()
 
+    COMMAND_SEQUENCE_TIMEOUT = 1.0
+
     def __init__(self, **kwargs):
         self.comms = kwargs.get('comms', self.comms)
         self._running = Event()
@@ -236,7 +238,7 @@ class RcpApi:
         while self._running.is_set():
             try:
                 # Block for 1 second for messages
-                command = self._command_queue.get(True, 1.0)
+                command = self._command_queue.get(True, RcpApi.COMMAND_SEQUENCE_TIMEOUT)
 
                 command_list = command.command_list
                 rootName = command.rootName
--- a/autosportlabs/racecapture/api/rcpapi.py
+++ b/autosportlabs/racecapture/api/rcpapi.py
@@ -202,7 +202,7 @@ class RcpApi:
                 else:
                     sleep(0.25)
 
-        Logger.info("RCPAPI: RxWorker exiting")
+        Logger.info("RCPAPI: msg_rx_worker exiting")
 
     def rcpCmdComplete(self, msgReply):
         self.cmdSequenceQueue.put(msgReply)
--- a/autosportlabs/racecapture/api/rcpapi.py
+++ b/autosportlabs/racecapture/api/rcpapi.py
@@ -114,6 +114,7 @@ class RcpApi:
 
     def cleanup_comms(self):
         self.comms.cleanup()
+        self.stop_message_rx_worker()
 
     def shutdown_comms(self):
         Logger.info('RCPAPI: shutting down comms')
--- a/autosportlabs/racecapture/api/rcpapi.py
+++ b/autosportlabs/racecapture/api/rcpapi.py
@@ -94,7 +94,6 @@ class RcpApi:
     def _start_message_rx_worker(self):
         self._running.set()
         t = Thread(target=self.msg_rx_worker)
-        t.daemon = True
         t.start()
         self._msg_rx_thread = t
 
@@ -110,7 +109,6 @@ class RcpApi:
 
     def _start_cmd_sequence_worker(self):
         t = Thread(target=self.cmd_sequence_worker)
-        t.daemon = True
         t.start()
         self._cmd_sequence_thread = t
 
@@ -631,7 +629,6 @@ class RcpApi:
     def start_auto_detect_worker(self):
         self._auto_detect_event.clear()
         t = Thread(target=self.auto_detect_worker)
-        t.daemon = True
         t.start()
         self._auto_detect_worker = t
 
--- a/autosportlabs/racecapture/api/rcpapi.py
+++ b/autosportlabs/racecapture/api/rcpapi.py
@@ -11,29 +11,29 @@ from kivy.clock import Clock
 from kivy.logger import Logger
 from traceback import print_stack
 
-TRACK_ADD_MODE_IN_PROGRESS      = 1
-TRACK_ADD_MODE_COMPLETE         = 2
+TRACK_ADD_MODE_IN_PROGRESS = 1
+TRACK_ADD_MODE_COMPLETE = 2
 
-SCRIPT_ADD_MODE_IN_PROGRESS     = 1
-SCRIPT_ADD_MODE_COMPLETE        = 2
+SCRIPT_ADD_MODE_IN_PROGRESS = 1
+SCRIPT_ADD_MODE_COMPLETE = 2
 
-DEFAULT_LEVEL2_RETRIES          = 3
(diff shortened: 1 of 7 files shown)


autosportlabs/racecapture/api/rcpapi.py
```diff
@@ -114,6 +114,7 @@ def init_comms(self, comms):
 
     def cleanup_comms(self):
```
stieg: So what is the difference between `cleanup_comms` and `shutdown_comms` exactly?  I'm just trying to figure out if line 117 is the correct place to put this?  The location is confusing since the order in which these methods should be called isn't entirely clear.  I assume it would go `shutdown_comms` then `cleanup_comms` but that is a guess.  Will you clarify that please/

brentpicasso: shutdown_comms is only used by the firmware update screen and only shuts down the rc_api message loop, so it won't try to grab and process data while the firmware is updating.   There are other daemon threads that remain running in the background. 

shutdown_comms could be renamed something more accurate. will investigate.

brentpicasso: Well, further digging, the function name is technically correct. The firmware update screen disables auto-connection recovery, *then* shuts down any comm port activity. The rc_api idles until the firmware update screen re-enables it. The message loop thread *does* remain alive, it just pauses waiting for data.

A better set of functions might be "pause_api" and "resume_api".   Also place the other daemon threads under control so they can be explicitly exited like rc_api thread.

stieg: Sounds like a state machine implementation there.  So yeah those method names sound much more descriptive.

autosportlabs/racecapture/api/rcpapi.py
```diff
@@ -91,29 +91,38 @@ def recover_connection(self):
 
     def _start_message_rx_worker(self):
         self._running.set()
-        self._msg_rx_thread = Thread(target=self.msg_rx_worker)
-        self._msg_rx_thread.daemon = True
-        self._msg_rx_thread.start()
+        t = Thread(target=self.msg_rx_worker)
+        t.daemon = True
```
stieg: If you are having this issue both the [Referenced Stackoverflow Question](http://stackoverflow.com/questions/4079810/python-thread-exception-errors-when-exit-the-whole-program) and [Python threading Docs](https://docs.python.org/2/library/threading.html) mention not to set the Daemonic flag.  Is there a reason you are still setting it?

brentpicasso: I left it enabled b/c harmless and it's a backup in case we've failed to terminate it cleanly, somehow. As a daemon thread the python runtime will terminate it regardless.

autosportlabs/racecapture/api/rcpapi.py
```diff
@@ -91,29 +91,38 @@ def recover_connection(self):
...
+        t.start()
+        self._msg_rx_thread = t
 
-    def stop_message_rx_worker(self):
+    def _shutdown_workers(self):
         Logger.info('RCPAPI: Stopping msg rx worker')
         self._running.clear()
+        # this allows the auto detect worker to fall through if needed
+        self._auto_detect_event.set()
```
stieg: Would it be wiser to set these variables _after_ you have joined the threads?

autosportlabs/racecapture/api/rcpapi.py
```diff
@@ -91,29 +91,38 @@ def recover_connection(self):
...
+    def init_api(self, comms):
         self.comms = comms
         self._start_message_rx_worker()
         self._start_cmd_sequence_worker()
         self.start_auto_detect_worker()
         Clock.schedule_interval(lambda dt: comms.keep_alive(), COMMS_KEEP_ALIVE_TIMEOUT)
 
-    def cleanup_comms(self):
-        self.comms.cleanup()
+    def shutdown_api(self):
```
stieg: Nice to see things getting more consistent :)

autosportlabs/racecapture/api/rcpapi.py
brentpicasso: It's specifically this way - it triggers an auto-detect so it can fall through and be caught, and then exited. It's a bit of a hack, but works.

autosportlabs/racecapture/api/rcpapi.py
```diff
@@ -223,9 +232,11 @@ def _queue_multiple(self, command_list, root_name, win_callback, fail_callback):
         self._command_queue.put(command)
 
     def cmd_sequence_worker(self):
-        while True:
+        Logger.info('RCPAPI: cmd_sequence_worker starting')
+        while self._running.is_set():
             try:
-                command = self._command_queue.get() #this blocks forever
+                # Block for 1 second for messages
+                command = self._command_queue.get(True, 1.0)
```
stieg: Could set this to something smaller like 0.5 to make quit more responsive.  Perhaps define a magic value at the top and use it for consistency?

brentpicasso: Yes - need to fix the magic number, but it's only needed when exiting.

autosportlabs/racecapture/api/rcpapi.py
```diff
@@ -625,16 +645,21 @@ def on_ver_win(value):
...
-        while True:
+        while self._running.is_set():
             try:
                 self._auto_detect_event.wait()
                 self._auto_detect_event.clear()
                 self._enable_autodetect.wait()
+                # check again if we're shutting down
+                # to prevent a needless re-detection attempt
+                if not self._running.is_set():
+                    break
```
stieg: Seems strange to have a duplicate check to the while here.  Would re-arranging the order fix this?

autosportlabs/racecapture/api/rcpapi.py
```diff
@@ -690,8 +715,11 @@ def on_ver_win(value):
                 Logger.error('RCPAPI: Error running auto detect: ' + str(e))
                 Logger.debug(traceback.format_exc())
             finally:
-                Logger.info("RCPAPI: auto detect finished. port=" + str(comms.port))
-                self._auto_detect_busy.clear()
-                self.removeListener("ver", on_ver_win)
-                self.sendCommandLock.release()
-                sleep(AUTODETECT_COOLOFF_TIME)
+                if self._running.is_set():
```
stieg: Why is this needed?  Why can't the program just cleanup like it supposed to?

autosportlabs/racecapture/databus/databus.py
```diff
@@ -193,6 +192,13 @@ def startDataPump(self, data_bus, rc_api):
             # we're already running, refresh channel meta data
             self.meta_is_stale()
 
+    def stop(self):
+        self._running.clear()
+        try:
+            self._sample_thread.join()
```
stieg: Strange that this would fail.  Any reason?

autosportlabs/racecapture/api/rcpapi.py
stieg: I would argue that if we are failing to terminate cleanly we should fail and fix it.  Its a programming error at that point.  I argue we should nuke this flag and fix it properly so we exit like a well behaved program.

autosportlabs/racecapture/api/rcpapi.py
brentpicasso: it's to cause the auto-detection to fall through immediately if we're self._running is cleared (signal to not run anymore).

Otherwise it'll needlessly trigger a re-detection effort, which will cause noise and delays in the app exiting.

autosportlabs/racecapture/api/rcpapi.py
brentpicasso: It was needed b/c when it would fall through due to the break it would leave things in an unexpected state.  However, I moved the try block down after the break so I wouldn't need to deal with this situation; so, it's cleaner.

autosportlabs/racecapture/api/rcpapi.py
brentpicasso: There is a small chance where there's a race condition when we shut down the app we try to join a thread that doesn't exist yet, *then* it creates the thread immediately afterwards. 

If the thread isn't daemonic, there would be left zombie threads lying around in people's task manager.  The daemon thread is a backup to handle this rare case.

autosportlabs/racecapture/databus/databus.py
brentpicasso: Mostly in case we try to exit *before* the thread is created. 

It's more pythonic to ask for forgiveness than to check if something is possible.  See http://stackoverflow.com/questions/7604636/better-to-try-something-and-catch-the-exception-or-test-if-its-possible-first

autosportlabs/racecapture/api/rcpapi.py
stieg: Huh?  I figure it would be that case if it _was_ a deamonic thread.  If it was a child thread without a parent then init would reap it unless it was treated as a deamon.

brentpicasso: Back to the original issue - daemon threads were fine, but they were sometimes generating otherwise harmless exceptions upon application shut down. These shut-down exceptions were showing up on Sentry in a high volume. 

So this is all in an effort to get threads to shut down cleanly, to minimize those harmless exceptions.   I had thought of squelching those exceptions in the crash handler, but thought attempting to close them down cleanly first would be the better design.

stieg: > I had thought of squelching those exceptions in the crash handler, but thought attempting to close them down cleanly first would be the better design.

I agree.  This is the proper approach.  So here is the line that is concerning to me from the Python docs:

> Daemon threads are abruptly stopped at shutdown. Their resources (such as open files, database transactions, etc.) may not be released properly. If you want your threads to stop gracefully, make them non-daemonic and use a suitable signalling mechanism such as an Event.

This is why I am persistent about this.  Because this very sentence seems to counter what you are trying to do.  Perhaps I am mistaken.

brentpicasso: We are attempting to shut them down before python shuts down at the very end - called from the on_stop() in main.py 

If for some reason they don't get stopped cleanly, the python runtime will come through later, at the very end and do a final cleanup. 

If they are abruptly stopped there may be a chance that a resource the thread was dependent upon would go away, causing those harmless but noisy exceptions to show up in the Sentry logs.

PR #881, opened 2016-02-04
Resize graph if plots are added or removed

### LINKED ISSUE #855: Graph doesn't resize when you deselect laps in analysis view
Graph doesn't resize when you unload a lap that had a lot longer distance than the rest: http://i.imgur.com/Ln66c1f.png

brentpicasso:
Resolves #855


ryandoherty:
:shipit:

PR #891, opened 2016-02-19
Upgrade path for track ids being returned as integers instead of strings

ryandoherty:
So, someone changed the RCL api to return integers instead of strings for track ids, which breaks all sorts of things in the app. 

This PR:

1. Detects we haven't fixed this issue using a preference that defaults to False, then is set to True once tracks are updated. 
2. Prompts the user they need to update their track maps 
3. Refresh tracks to use the new integer id

To test, make a backup of the venues folder in the app data dir first (these are the old string based ids). Put it back and update the preferences.ini file and set `integer_map_ids` to `False` to repeat the test.


stieg:
So does this imply that all apps in the field now are currently broken?

ryandoherty:
Somewhat, they can't update their track list. If they've previously downloaded tracks the app works fine.

stieg:
Sounds like a moderate issue then since we have to assume people have not updated their apps (because people don't update their apps no matter how much we ask them too).

ryandoherty:
Discussed in chat, going to 'fix' the api to behave as before and return strings.

PR #893, opened 2016-02-19
Demote/Promote various logging messages to quiet down terminal logging

stieg:
Currently we print out too many info level messages that really aren't
all that helpful.  This change demotes some of those messages and
promotes others that should be at a higher level (like warning).

Result is a much nicer terminal in normal usage, and a nicely
debuggable terminal when the app is fired up with the -d flag.


brentpicasso:
:ship:

Issue #888, opened 2016-02-13
Start BT device search by the RaceCapture name


kosenko-max:
Due to fix 
* Connect to the first RaceCapture device it can detect, regardless of the name provided
it's now going through all unrelated and disconnected devices. It would be nice if it can save 3.5 minutes time and try original name first...

brentpicasso:
Thanks. Optimizing this would be helpful.  What we can do is:

1- first attempt; scan all names for the corrected RaceCapture device.
2- upon successful connection, save the last successful name in preferences 
3- upon app re-start, try connecting to the saved name first
4- if app cannot connect to the saved name, fall back to searching all the names.  (goto 1 above)

Also worth noting, this is only showing up in the beta, not production release. Back-story: This is to ultimately support allowing people to customize the racecapture bluetooth name.

PR #898, opened 2016-03-05
P/brentpicasso/remember last known port

brentpicasso:
Update to re-connect to the last known port for RCP. If not found on the last known port, fall back to searching all available ports.

### DIFF
--- a/autosportlabs/racecapture/api/rcpapi.py
+++ b/autosportlabs/racecapture/api/rcpapi.py
@@ -675,7 +675,10 @@ class RcpApi:
                     if last_known_port:
                         Logger.info('RCPAPI: trying last known port first: {}'.format(last_known_port))
                         # ensure we remove it from the existing list
-                        ports.remove(last_known_port)
+                        try:
+                            ports.remove(last_known_port)
+                        except ValueError:
+                            pass
                         ports = [last_known_port] + ports
                     Logger.debug('RCPAPI: Searching for device on all ports')
 
--- a/autosportlabs/racecapture/api/rcpapi.py
+++ b/autosportlabs/racecapture/api/rcpapi.py
@@ -67,13 +67,14 @@ class RcpApi:
 
     COMMAND_SEQUENCE_TIMEOUT = 1.0
 
-    def __init__(self, **kwargs):
+    def __init__(self, settings, on_disconnect, **kwargs):
         self.comms = kwargs.get('comms', self.comms)
         self._running = Event()
         self._running.clear()
         self._enable_autodetect = Event()
         self._enable_autodetect.set()
-        self._disconnect_callback = kwargs.get('on_disconnect')
+        self._settings = settings
+        self._disconnect_callback = on_disconnect
 
     def enable_autorecover(self):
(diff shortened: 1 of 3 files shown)


autosportlabs/racecapture/api/rcpapi.py
```diff
@@ -669,6 +670,13 @@ def on_ver_win(value):
                     ports = [comms.port]
                 else:
                     ports = comms.get_available_ports()
+                    last_known_port = self._settings.userPrefs.get_pref('preferences', 'last_known_port')
```
brentpicasso: To avoid repeating the section / key value, I was thinking of putting a method or property in userPrefs that wraps these keys, or create a  local method that wraps these keys. thoughts?

brentpicasso: at the same time, IDK if userPrefs should have knowledge of another part of the system; separation of concerns and all that.

stieg: A wrapper method to help out probably isnt bad if it saves duplication

ryandoherty: Part of me feels like the RCP API shouldn't need a prefs object to work, something should just pass in what port(s) to check first. Then when it does connect, it could fire an event with the port it used. Makes it easier to write tests and easier to re-use this class in other projects or tools.

ryandoherty:
Overall looks good minus my thought about removing RcpApi's knowledge of a preferences object.

autosportlabs/racecapture/api/rcpapi.py
brentpicasso: Yeah, I get that. We pass settings/preferences to many things in the systems as a dependency with a standard interface. I think because it is passed in during construction it is specified as a dependency, it's generally OK.  

What would be bad is if it was accessing a static method of a class off to the side, cutting across layers of architecture

ryandoherty: Just because we do it elsewhere doesn't mean it's a good idea. It's tightly coupling multiple independent pieces of code together, making reuse more difficult. I vote to remove the dependency. Andrew can cast the swing vote :P

ryandoherty: CHINA

brentpicasso: This isn't an anti-pattern. Passing in config objects, database objects, and other bits of dependencies is a common thing and is basis of IoC and dependency injection.  Passing in the 'system config object' falls into a common expression of this pattern.

Please see: http://stackoverflow.com/questions/3058/what-is-inversion-of-control and
https://en.wikipedia.org/wiki/Inversion_of_control

ryandoherty: Good point. I still think that for this particular object, the RCPApi, it would benefit from not needing a settings/userprefs object so the RCPApi object could potentially be reused elsewhere in a different project. I only say that strongly for this object because I can see it being extracted at some point into a separate lib or Python package.

But anyway, :shipit:

brentpicasso:
Resolves #888

Issue #902, opened 2016-03-18
Initial downloading of track maps is slow if BT is not connected (Android)


brentpicasso:
If BT is not connected initial trackmap downloading seems slow - or may take a very long time to complete.

Issue #468, opened 2015-10-26
bundle existing tracks with installer to prevent a first time download


ryandoherty:

brentpicasso:
resolved in 1.4.4.

PR #903, opened 2016-03-20
pre-package trackmaps with app; add timeout to trackmap downloading t…

brentpicasso:
Improve first time experience for users by pre-packaging track maps as defaults. 

* If no trackmaps are present, defaults are loaded. 
* Added timeout for url open to when fetching trackmap updates
* removed first time download prompt
* naming refactoring of TrackMap class UI widget
* formatting / other cleanup refactoring

addresses #902 #468

### DIFF
--- a/autosportlabs/racecapture/tracks/trackmanager.py
+++ b/autosportlabs/racecapture/tracks/trackmanager.py
@@ -3,16 +3,20 @@ import time
 import copy
 import errno
 import string
-import logging
 from threading import Thread, Lock
 import urllib2
 import os
 import traceback
+from StringIO import StringIO
+import gzip
+import zipfile
 from autosportlabs.racecapture.geo.geopoint import GeoPoint, Region
+from autosportlabs.util.timeutil import time_to_epoch
 from kivy.logger import Logger
-from utils import time_to_epoch
+
 TRACK_DEFAULT_SEARCH_RADIUS_METERS = 2000
 TRACK_DEFAULT_SEARCH_BEARING_DEGREES = 360
+TRACK_DOWNLOAD_TIMEOUT = 30
 
 class TrackMap:
     """Very generic object wrapper around RCL's API endpoint for venues
@@ -77,7 +81,7 @@ class TrackMap:
 
         if not self.short_id > 0:
             raise Warning("Could not parse trackMap: short_id is invalid")
-    
+
     def to_dict(self):
         """Create a dict for saving this object's data. Usually for saving to a file
         """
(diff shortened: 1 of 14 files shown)


autosportlabs/racecapture/tracks/trackmanager.py
```diff
@@ -309,11 +319,21 @@ def load_current_tracks_worker(self, success_cb, fail_cb, progress_cb=None):
...
             self.update_lock.release()
-        
+
+    def check_load_default_tracks(self):
+        track_file_names = os.listdir(self.tracks_user_dir)
+        if (len(track_file_names) == 0):
+            Logger.info("TrackManager: No tracks found; loading defaults")
+            try:
+                with zipfile.ZipFile(os.path.join(self.base_dir, 'defaults', 'default_tracks.zip'), 'r') as z:
+                    z.extractall(self.tracks_user_dir)
```
ryandoherty: Have we tested how long this takes on a slow tablet/phone? Just want to make sure we don't lock up things for a long time.

build_release_apk.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 set -e
 rm -rf bin
+python -m build_tools.build_default_tracks
```
ryandoherty: Need any sort of checking here if this fails? Do we want to fail the build completely?

autosportlabs/racecapture/tracks/trackmanager.py
brentpicasso: On the crappy Asus reference tablet it's about a quarter second, so it should be negligible for a long time.

build_release_apk.sh
brentpicasso: conveniently, nope.  If the python script raises an error, the bash script stops.

ryandoherty: :+1:

autosportlabs/racecapture/tracks/trackmanager.py
ryandoherty: :+1:

defaults/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Ignore everything in this directory
+*
+# Except this file
+!.gitignore
```
ryandoherty: I'm confused why we need this gitignore. Can we use the top level one?

brentpicasso: it's to force the presence of the directory, and prevent anything from being checked in to this directory.

ryandoherty:
Overall looks good and works great! A great improvement.

Issue #884, opened 2016-02-09
Some GPS, Lap Timing data shown on dashboard even when it's switched off


kosenko-max:
Hi.

In the app I switched off GPS, LapTimer and Accelerometer, but I still see some of them on dashboard. Since some of them disappeared - I guess it's a bug.
![screen shot 2016-02-09 at 16 26 27](https://cloud.githubusercontent.com/assets/468956/12911094/df42d1ec-cf4a-11e5-99b5-d4aa677bccd2.png)

stieg:
Would seem like it?  What happens if you restart the app?  Did all the same channels re-appear?  May be related to internal meta-channels (calculations we do app side).

brentpicasso:
Yes, this is an issue as these additional channels are virtual / computed channels generated by the app.  The app should not create these channels if the dependent channels are not present in the stream.

brentpicasso:
resolved in 1.4.4

PR #904, opened 2016-03-20
if lap timing is off, don't create the bestlap / lapdelta channels

brentpicasso:
Addresses #884 

Introspect incoming channel meta to see if laptime is present; if so; add the related best lap and lap delta filter channels


ryandoherty:
Looks good, :shipit:

PR #908, opened 2016-03-27
remove selected laps when session is deleted

### LINKED ISSUE #800: Track map and graph not cleared when sessions are removed
STR:
1) Import log file with laps
2) Select a lap
3) Remove session (big red 'X')
4) See lap data in map and chart

![]
(http://i.imgur.com/Oaz5bg5.png)

brentpicasso:
If a session is deleted, remove any laps that might be currently selected. Resolves #800

### DIFF
--- a/autosportlabs/racecapture/views/analysis/sessionbrowser.py
+++ b/autosportlabs/racecapture/views/analysis/sessionbrowser.py
@@ -139,7 +139,7 @@ class SessionBrowser(AnchorLayout):
 
     def __init__(self, **kwargs):
         super(SessionBrowser, self).__init__(**kwargs)
-        self.register_event_type('on_lap_selected')
+        self.register_event_type('on_lap_selection')
         accordion = Accordion(orientation='vertical', size_hint=(1.0, None))
         sv = ScrollContainer(size_hint=(1.0, 1.0), do_scroll_x=False)
         self.selected_laps = {}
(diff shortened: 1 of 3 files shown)


autosportlabs/racecapture/views/analysis/sessionbrowser.py
```diff
@@ -225,29 +240,47 @@ def append_lap(self, session, lap, laptime):
...
+
+    def _notify_lap_selected(self, source_ref, selected):
+        '''
+        Deselect all laps specified in the list of source refs
+        :param source_ref the source reference for the lap
+        :type SourceRef
+        :param selected true if the lap is selected
+        :type Boolean 
+        '''
+        self.dispatch('on_lap_selected', source_ref, selected)
```
ryandoherty: So the `lap_selected` event is being used to mean both when a lap is selected and when it's unselected, depending on the data published in the event. Seems like this could be confusing to someone who subscribes to this event. I think we should have a `lap_deselected` event instead. Thoughts?

brentpicasso: I was extending the metaphor used by the widget itself where a single selection event propagated the the selection state. It's more of a state change. will think of maybe a better name for the event

ryandoherty: That sounds good to me :+1:

brentpicasso: Thanks. new event name in this commit: https://github.com/autosportlabs/RaceCapture_App/pull/908/commits/e232d8c0c22ed9ed701d3c7a030aa474cbb8b819

ryandoherty:
:shipit:

Issue #862, opened 2016-01-30
Channel stats infrastructure


brentpicasso:
compute and store channel statistics upon import. 

compute min, max, average and std. deviation for channels and laps.

PR #910, opened 2016-03-28
switch lap querying to use CurrentLap and ElapsedTime throughout

### LINKED ISSUE #863: Base lap / session view on currentLap and elapsed time
Related to #862 - once we have stats computed and stored we'll be able to display laps based on calculated stats data, not query across the main data point set.

brentpicasso:
We should be selecting laps based on CurrentLap channel and display the lap time as the max of the ElapsedTime for that CurrentLap, and not use LapCount / LapTime channels, as those reference the number of laps so far, and the last lap time.

This Resolves #863

### DIFF
--- a/autosportlabs/racecapture/datastore/datastore.py
+++ b/autosportlabs/racecapture/datastore/datastore.py
@@ -239,7 +239,7 @@ class DatalogChannel(object):
         return self.name
 
 class DataStore(object):
-    EXTRA_INDEX_CHANNELS = ["LapCount","CurrentLap"]    
+    EXTRA_INDEX_CHANNELS = ["CurrentLap"]    
     val_filters = ['lt', 'gt', 'eq', 'lt_eq', 'gt_eq']
     def __init__(self):
         self._channels = []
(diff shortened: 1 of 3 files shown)


autosportlabs/racecapture/datastore/datastore.py
```diff
@@ -892,7 +899,31 @@ def get_sessions(self):
...
+        '''
+        Fetches a list of lap information for the specified session id
+        :param session_id the session id
+        :type session_id int
+        :returns list of Lap objects
+        :type list 
+        '''        
+        laps = []
+        c = self._conn.cursor()
+        for row in c.execute('''SELECT DISTINCT sample.session_id AS session_id,
```
ryandoherty: This query is going to become slower and slower as the table grows, we should prioritize generating lap data on import. Not a blocker for this PR, but certainly for GA for analysis.

ryandoherty:
Seems fine aside from the potentially slow query.

autosportlabs/racecapture/datastore/datastore.py
brentpicasso: Right. #862 makes a reference to #863 about pre-computing stats.

PR #913, opened 2016-03-30
Change race-capture.com references to Podium

ryandoherty:
Pretty simple. This already works because podium.live is the RCL codebase.


brentpicasso:
looks good. merging.

PR #915, opened 2016-04-02
Update to Kivy 1.9.1

brentpicasso:
Updates to Kivy 1.9.1

Some quirks with 1.9.1 - touch behavior for buttons in scroll containers needed to be tweaked.  Without the tweaks it was hard to get buttons and text fields to activate; kivy would first register them as scroll events. The tweaks optimize the touch behavior so it's easier to touch the widgets.

The code input window was referencing a monospaced font that was no longer present. Added our own monospaced font and referenced that.


autosportlabs/racecapture/menu/homepageview.py
```diff
@@ -1,5 +1,5 @@
 import kivy
-kivy.require('1.9.0')
+kivy.require('1.9.1')
```
stieg: We would be smart to centralize this somewhere.  Or make the `1.9.1` a string that gets set in a common location.

autosportlabs/racecapture/views/configuration/rcp/scriptview.kv
```diff
@@ -17,7 +17,7 @@
                 LuaCodeInput:
                 	rcid: 'script'
                     size_hint_y: None
-                    font_name: 'data/fonts/DroidSansMono.ttf'
+                    font_name: 'resource/fonts/RobotoMono-Regular.ttf'
```
stieg: So all fonts moved or you just changed this one or?  Just out of curiosity.

stieg:
Changes look reasonably sane.  Build environment doesn't work for me right off the bat.  Fighting it a bit now.  Will see if I can get a working version and may post a PR against this branch or against the project if it merges in cleanly.  Try doing a brand new build environment on your end @brentpicasso (if feasable) and see if it works.

requirements.txt
```diff
@@ -1,5 +1,7 @@
 https://github.com/autosportlabs/ihexpy/archive/master.zip
 dependencies/asl_f4_loader-0.0.7.tar.gz
+kivy==1.9.1
+cython==0.23
```
stieg: Move this above kivy since kivy depends on cython and this is done in order (at least on my machine).

stieg: Actually... that didn't work either.  It seems one must install cython before attempting to install anything related to kivy.

autosportlabs/racecapture/menu/homepageview.py
brentpicasso: I don't think this a big concern. We'd then have to include some common file or module in order to bring in the string.

autosportlabs/racecapture/views/configuration/rcp/scriptview.kv
brentpicasso: kivy 1.9.1 eliminated DroidSansMono, so we have our own font, as described in the top PR message.

ryandoherty:
This all works on OSX, minor issue of that on OSX you can install Kivy separately, so if you do that, then run `kivy -m pip install -r requirements.txt`, you end up install Kivy inside Kivy :P

brentpicasso:
Thanks. Are the installation instructions good as is in Readme?

Brent Picasso
Autosport Labs

ryandoherty:
Not really, I'd use the instructions they have here: https://kivy.org/docs/installation/installation-osx.html

Installing Kivy stuff on OSX is an exercise in futility for many. Takes a long time and there's always some random failures (like having to use `sudo` for some things). I'd recommend telling folks to just install the prebuilt package, it's a lot easier. 

I haven't even tested if I can build a RC app installer yet, we'll see how that goes.

PR #917, opened 2016-04-02
do not ever ever start sample worker if on mobile platform

brentpicasso:
this acknowledges that on the mobile platform we should never, ever enable the polling sampler.  There could be a case (and it's been seen on slower hardware maybe) where the wait period times out and the app starts polling for samples over BT.  We should only poll on USB, and then later, we'll have the API to enable / disable streaming.  This is a step in that direction.

This may also help mitigate some of the timeouts seen when writing configurations if the sampleworker incorrectly starts up. still, there's likely more work needed in this area. 

Finally, this code is simpler. simpler = better.

### DIFF
--- a/autosportlabs/racecapture/databus/databus.py
+++ b/autosportlabs/racecapture/databus/databus.py
@@ -249,8 +249,8 @@ class DataBusPump(object):
                 rc_api.sample()
                 sleep(SAMPLE_POLL_INTERVAL_TIMEOUT)
             except Exception as e:
-                sleep(SAMPLE_POLL_EXCEPTION_RECOVERY)
                 Logger.error('DataBusPump: Exception in sample_worker: ' + str(e))
+                sleep(SAMPLE_POLL_EXCEPTION_RECOVERY)
 
         Logger.info('DataBusPump: sample_worker exiting')
         safe_thread_exit()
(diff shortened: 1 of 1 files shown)


autosportlabs/racecapture/databus/databus.py
```diff
@@ -239,29 +243,14 @@ def request_meta(self):
...
-                    Logger.error('DataBusPump: Exception in sample_worker: ' + str(e))
-                finally:
-                    sample_event.clear()
+        while self._running.is_set():
+            try:
+                rc_api.sample()
+                sleep(SAMPLE_POLL_INTERVAL_TIMEOUT)
+            except Exception as e:
+                sleep(SAMPLE_POLL_EXCEPTION_RECOVERY)
+                Logger.error('DataBusPump: Exception in sample_worker: ' + str(e))
```
stieg: Move this above the sleep.

stieg: Seems like the sleep is common and what only changes is the time.  Is there any major difference in the time values?

stieg:
Seems logical.  Minor comments.  Fix and then :shipit:

autosportlabs/racecapture/databus/databus.py
brentpicasso: K, fixing logger message ordering. 

the exception recovery is a different delay value (0.05 vs 10 seconds); if there was an error I didn't want those errors to occur at a high rate. The delay allows for a backoff period.

PR #921, opened 2016-04-04
Fix incorrect mouse position with high dpi screens in analysis view

ryandoherty:


autosportlabs/racecapture/views/analysis/linechart.py
```diff
@@ -205,7 +207,7 @@ def on_mouse_pos(self, x, pos):
         if not self.collide_point(pos[0], pos[1]):
             return False
         
-        self.dispatch_marker(pos[0], pos[1])
+        self.dispatch_marker(pos[0] * self.metrics_base.density, pos[1])
```
brentpicasso: We should handle both X and Y, just in case we care about Y

ryandoherty: Cool, done!

brentpicasso:
Thanks!

PR #925, opened 2016-04-05
Adjust center dashboard gauge

brentpicasso:
slightly shrink center dashboard gauge to eliminate / reduce chances of overlap on certain aspect ratio displays


ryandoherty:
:shipit:

PR #928, opened 2016-04-08
allow for telemetry id field blanking; general code cleanup

### LINKED ISSUE #927: Cannot blank device ID
If you erase a device ID field in the app it will not erase the value completely; the first character remains.

e.g. if the ID was 

123456

erasing it would leave '1' as the value after reading config back.

Verified caused by app

brentpicasso:
Resolves #927


ryandoherty:
tested, :shipit:

Issue #940, opened 2016-04-18
Simple pit stop timer


brentpicasso:
Enable a simple popup pit-stop timer.  When the race car comes into pit, pop up a window that starts counting up time.

* When car resumes, auto-dismiss popup
* A button on the popup allows the user to dismiss

CurrentLap should be > 0 so it doesn't get triggered in non racing mode.

PR #943, opened 2016-04-19
P/brentpicasso/pit stop timer

brentpicasso:
Adds an automatic pit stop timer that tries to be as smart as possible:

stop threshold speed = 2
resume threshold speed = 55
exit alert speed = 35

Pit stop timer will only activate if we're first in race mode (at least 55 / with current lap > 0) and then drop below stop threshold speed, to minimize false popups.

### DIFF
--- a/autosportlabs/racecapture/views/dashboard/widgets/stopwatch.py
+++ b/autosportlabs/racecapture/views/dashboard/widgets/stopwatch.py
@@ -166,7 +166,6 @@ class PitstopTimerView(BoxLayout):
             self._toggle_exit_speed_alert(ColorScheme.get_error())
         elif self.current_speed > self.exit_speed_alert_threshold * self._SPEED_ALERT_THRESHOLD:
             self.ids.speed_rect.rect_color = ColorScheme.get_alert()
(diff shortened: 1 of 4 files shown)


autosportlabs/racecapture/views/dashboard/widgets/stopwatch.py
```diff
@@ -0,0 +1,283 @@
...
+    _TIMER_WIDGET_SIZE_HINT = 0.4
+    _SPEED_ALERT_THRESHOLD = 0.85
+    _POPUP_SIZE_HINT = (0.75, 0.8)
+    #Settings
+    stop_threshold_speed = NumericProperty(2.0)
+    resume_threshold_speed = NumericProperty(55.0)
+    exit_speed_alert_threshold = NumericProperty(35.0)
+    
+    exit_speed_color = ListProperty(ColorScheme.get_dark_background_translucent())
+    title = StringProperty('')
```
ryandoherty: I see this variable up here, then it's an argument in __init__, which is assigned to self. Still needed?

ryandoherty:
Looks good! :shipit:

brentpicasso:
resolves #940

PR #968, opened 2016-05-02
ensure write button is activated when loading a config from file. 

### LINKED ISSUE #955: write config does not enable after loading a config
When loading a config the write button should be enabled so the config can be written to the unit. 

work around is to make a separate config change to enable the write button.

brentpicasso:
simplify code / remove scheduled calls. resolves #955


brentpicasso:
sorry about the whitespace, editor fixed things up. use ?w=1 to see what really changed.

ryandoherty:
:shipit:

Issue #905, opened 2016-03-21
Status screen messages truncated


ryandoherty:
Would be nice if the messages wrapped.

![unknown](https://cloud.githubusercontent.com/assets/45628/13927133/a5aca5bc-ef4c-11e5-993a-1c5c599640e9.png)

PR #971, opened 2016-05-06
Wrap text in status view

ryandoherty:
Text clips in the status view (#905), which makes it really hard to debug. This fixes it.

### DIFF
--- a/autosportlabs/racecapture/views/status/statusview.py
+++ b/autosportlabs/racecapture/views/status/statusview.py
@@ -27,7 +27,10 @@ class StatusTitle(StatusLabel):
     pass
 
 class StatusValue(StatusLabel):
-    pass
+
+    def __init__(self, **kwargs):
+        super(StatusLabel, self).__init__(**kwargs)
+        self.shorten = False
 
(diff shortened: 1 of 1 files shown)


autosportlabs/racecapture/views/status/statusview.py
```diff
@@ -27,7 +27,10 @@ class StatusTitle(StatusLabel):
     pass
 
 class StatusValue(StatusLabel):
-    pass
+
+    def __init__(self, **kwargs):
+        super(StatusLabel, self).__init__(**kwargs)
+        self.shorten = False
```
brentpicasso: Best is to define this in the statusview.kv file.

ryandoherty: That doesn't work, it's set in StatusLabel's init, so I have to override it here.

brentpicasso: gotcha. :shipit:

brentpicasso:
Can you merge this into 1.5.3 please instead of master?

ryandoherty:
Oops, will do!

PR #974, opened 2016-05-06
Add network status to status screen. Fix #887

### LINKED ISSUE #887: Add support for "state" key in Cell Net Status messages
As part of autosportlabs/RaceCapture-Pro_firmware/pull/416 there will be a new feature in the status message that provides a human readable string telling the user about the state of the cellular network.  It provides very useful message about your connectivity to the network (including informing you when the network is actively rejecting your SIM card).

This ticket tracks adding support for it in the app so it gets printed when present.

ryandoherty:


brentpicasso:
looks good :+1:

PR #975, opened 2016-05-06
Make dash nav arrows bigger for easier use. Fixes #965

### LINKED ISSUE #965: Make dashboard navigation arrows larger / easier to touch
Need to make it easier to navigate screens, especially with gloves.

Riffing off of the auto-brightening configuration icon, possibly make the icon size for the arrows grow larger if the screen is touched.  Or have them large and make them dim when not in use.

ryandoherty:
This change makes the arrows have a nearly 2x as large touch target while still fitting inside their container (afaict).


brentpicasso:
checked with regular scaling and ./emu_phone.sh; looks good. :shipit:

PR #976, opened 2016-05-09
Ios support

ryandoherty:
This PR is for bare minimum changes that allow our app to run on iOS. A few small changes to make the socket comms not throw errors (they don't actually do anything yet), build tools and modified theme file to make checkboxes actually show up.


autosportlabs/comms/socket/socketcomm.py
```diff
@@ -0,0 +1,46 @@
+from kivy.logger import Logger
```
stieg: Do we have a header for ASL .py files?  If not we should make one.

autosportlabs/comms/socket/socketcomm.py
```diff
@@ -0,0 +1,46 @@
...
+
+STAY_ALIVE_TIMEOUT = 4
```
stieg: NIT: s/STAY/KEEP/.  Also I assume this is in seconds.

install/ios/create_requirements.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/python
```
stieg: Do you all not have a `bin` directory for scripts?  Its what we do with firmware project.

install/ios/create_requirements.py
```diff
@@ -0,0 +1,38 @@
...
+
+"""
+iOS is complicated, we need to install some pip packages, but not all for various reasons
+This script takes our top level requirements.txt file and generates a new one with the
+blacklisted packages removed.
+"""
+
+import os
+local_dir = os.path.dirname(os.path.realpath(__file__))
+root_dir = os.path.dirname(os.path.join(local_dir, '../../'))
```
stieg: Dangerous b/c if filed is moved this breaks.  At a minimum make a comment about this.

stieg: Also no trailing slash should be needed.  Simply `../..`

install/ios/create_requirements.py
```diff
@@ -0,0 +1,38 @@
...
+ios_requirements_file_path = os.path.join(root_dir, 'ios_requirements.txt')
```
stieg: Wait... ios_requirements is part of .gitignore yet you are using it in this file too?  Seems wrong unless I missed something.

stieg: And why is this file at the root level?  I imagine b/c of something silly with XCode?

install/ios/create_requirements.py
```diff
@@ -0,0 +1,38 @@
...
+
+if os.path.isfile(ios_requirements_file_path):
+    os.remove(ios_requirements_file_path)
+
+exclude_list = open(os.path.join(local_dir, 'ios_pip_exclude.txt'), 'r').read().split('\n')
+req_list = open(os.path.join(root_dir, 'requirements.txt'), 'r').read().split('\n')
+
+exclude_list = [name for name in exclude_list if len(name) > 0]
```
stieg: Never seen this syntax before.  Interesting.

install/ios/create_requirements.py
```diff
@@ -0,0 +1,38 @@
...
+            break
+
+    return include
+
+ios_requirements_file = open(ios_requirements_file_path, 'w')
+
+
+include_list = filter(included, req_list)
+
+ios_requirements_file.write("\n".join(include_list))
```
stieg: This will always prepend a newline at the beginning (causing the first line to be empty).  Don't you want to always append instead?

requirements.txt
```diff
@@ -8,4 +8,4 @@ requests==2.2.1
 wsgiref==0.1.2
 pygments==2.0.2
 mock==1.3.0
-raven==5.5.0
+raven==5.5.0
```
stieg: whitespace change?

install/ios/create_requirements.py
ryandoherty: No we don't. I'd like to keep the platform installation tools separate, some platforms require a lot of specialized scripts or setup (OSX & iOS).

install/ios/create_requirements.py
ryandoherty: Is there any way for this script to figure out the root directory for this project in a way that won't break if it's moved?

install/ios/create_requirements.py
ryandoherty: It's a temporary file that is created as part of the build steps in `create_requirements.py`. XCode does a `pip install -r ios_requirements.txt --force-reinstall --ignore-installed --target .` to install the pip packages this app needs locally. Kivy on iOS is a freakin' mess.

install/ios/create_requirements.py
ryandoherty: Python list comprehensions FTW!

install/ios/create_requirements.py
ryandoherty: Hmm, the file doesn't have a newline at the beginning, just checked. Python is weird like that.

requirements.txt
ryandoherty: Dunno what happened, will check.

autosportlabs/comms/socket/socketcomm.py
ryandoherty: 👍

install/ios/create_requirements.py
stieg: Sure.  You could look for the existence of the .git directory which is always at the top level of any git project.  But that may be more complicated.  If you don't think the file will ever move, just make a nice big comment about it so that people that actually do their homework will see it.  If you think it will move recursively go up your tree until you see the .git directory and stop there.  You will have found the project root.

autosportlabs/comms/socket/socketcomm.py
ryandoherty: Nope, we can do that, will open an issue.

autosportlabs/comms/socket/socketcomm.py
ryandoherty: I'll change this in the ios-socket branch, it uses this variable in a few places.

requirements.txt
ryandoherty: Yup, whitespace. One blank character at the end of the line.

ryandoherty:
@stieg Comments responded to and/or code updated per your feedback.

stieg:
:shipit:

autosportlabs/comms/socket/socketcomm.py
```diff
@@ -0,0 +1,46 @@
+from kivy.logger import Logger
+
+KEEP_ALIVE_TIMEOUT_S = 4
+COMMAND_CLOSE = 'CLOSE'
+COMMAND_KEEP_ALIVE = 'PING'
+
+class SocketComm():
```
brentpicasso: should extend object

autosportlabs/comms/socket/socketcomm.py
```diff
@@ -0,0 +1,46 @@
...
+    CONNECT_TIMEOUT = 1.0
+    DEFAULT_TIMEOUT = 1.0
+    QUEUE_FULL_TIMEOUT = 1.0
+    _timeout = DEFAULT_TIMEOUT
+    port = None
```
brentpicasso: is this settable from outside the class? if not, pls use the _ prefix.

autosportlabs/comms/socket/socketcomm.py
ryandoherty: This was just enough to get the app running, all socket stuff is in a separate branch. I can address in the ios-socket branch.

brentpicasso: ok

brentpicasso:
nothing more to add on top of existing comments. looks good. :shipit:

autosportlabs/comms/socket/socketcomm.py
ryandoherty: Changed in ios-socket branch, will update there if applicable.

PR #977, opened 2016-05-09
[RFC] Wifi socket support

ryandoherty:
Initial PR for socket support for our app. This is requesting to merge into ios-support so only the changes for sockets are shown. Do not merge!

Important files are:

* socketcomm.py
* socketconnection.py

Other changes are mostly just renaming `port` to `device`.

Socket communication is nearly identical to serial communication. Separate threads read and write data to the socket and put/pull data from shared queues. Socket connection continually watches for data from the socket unless a shared event object is false, then it returns and ends.

Todo:
* tests


autosportlabs/comms/commsfactory.py
```diff
@@ -8,13 +8,17 @@
...
         from autosportlabs.comms.socket.socketcomm import SocketComm
-        return SocketComm(port=port, connection=SocketConnection())
+        return SocketComm(SocketConnection(), device)
     else:
-        from autosportlabs.comms.comms import Comms
-        return Comms(port=port, connection=SerialConnection())
+        from autosportlabs.comms.socket.socketconnection import SocketConnection
+        from autosportlabs.comms.socket.socketcomm import SocketComm
+        return SocketComm(SocketConnection(), device)
+        # from autosportlabs.comms.comms import Comms
```
brentpicasso: commented out code

autosportlabs/comms/socket/socketcomm.py
```diff
@@ -1,46 +1,158 @@
...
+from time import sleep
+import traceback
+from autosportlabs.comms.commscommon import PortNotOpenException
 
 STAY_ALIVE_TIMEOUT = 4
 COMMAND_CLOSE = 'CLOSE'
 COMMAND_KEEP_ALIVE = 'PING'
 
+
+def connection_thread_message_reader(rx_queue, connection, should_run):
```
brentpicasso: Recommend making these functions part of the SocketComm class, maybe mark them as private methods (_)

brentpicasso:
please add class and method comments per contributing doc using http://docutils.sourceforge.net/rst.html format

autosportlabs/comms/socket/socketcomm.py
```diff
@@ -1,46 +1,158 @@
...
-        self._connection = kwargs.get('connection')
+    def __init__(self, connection, device):
+        self.device = device
+        self._connection = connection
         Logger.info('SocketComm: init')
 
     def start_connection_process(self):
-        pass
+        rx_queue = Queue.Queue()
+        tx_queue = Queue.Queue(5)
```
brentpicasso: extract magic number; applies throughout

autosportlabs/comms/socket/socketcomm.py
```diff
@@ -1,46 +1,158 @@
...
+def connection_thread_message_reader(rx_queue, connection, should_run):
+    Logger.debug('SocketComm: connection thread message reader started')
+
+    while should_run.is_set():
+        try:
+            msg = connection.read(should_run)
+            if msg:
+                rx_queue.put(msg)
+        except:
+            Logger.error('SocketComm: Exception in connection_process_message_reader')
```
brentpicasso: looks like an exception in either thread will cause both to exit. intentional?

Also, do we want to exit on all exceptions? Quite possibly, but the connection might throw an exception that could be  benign enough to not warrant closing everything down possibly. thoughts?

autosportlabs/comms/socket/socketconnection.py
```diff
@@ -1,31 +1,79 @@
...
+
+        while keep_reading.is_set():
+            try:
+                data = self.socket.recv(4096)
+                if data == '':
+                    return None
+
+                msg += data
+
+                if msg[-2:] == '\r\n':
```
brentpicasso: what if we only get one character? will -2 generate an array index out of bound exception?

autosportlabs/racecapture/api/rcpapi.py
```diff
@@ -666,28 +666,29 @@ def on_ver_win(value):
...
+                if comms.device:
+                    devices = [comms.device]
                 else:
-                    ports = comms.get_available_ports()
-                    last_known_port = self._settings.userPrefs.get_pref('preferences', 'last_known_port')
-                    # if there was a last known port try this one first.
-                    if last_known_port:
-                        Logger.info('RCPAPI: trying last known port first: {}'.format(last_known_port))
+                    devices = comms.get_available_devices()
+                    # Backwards compatibility when fetching port, convert to device
```
brentpicasso: i think it would be safe to convert this preferences key to the name 'device' . the impact is that the value won't be set, and it'll have to do an initial search again.

autosportlabs/racecapture/databus/databus.py
```diff
@@ -190,11 +190,11 @@ def start(self, data_bus, rc_api):
         rc_api.addListener('s', self.on_sample)
         rc_api.addListener('meta', self.on_meta)
         self._running.set()
-        if not is_mobile_platform():
-            #only start the worker on desktop mode
-            t = Thread(target=self.sample_worker)
-            t.start()
-            self._sample_thread = t
+        # if not is_mobile_platform():
```
brentpicasso: left over commented code?

brentpicasso:
all of this is exciting to see.

autosportlabs/comms/socket/socketcomm.py
ryandoherty: I'm not sure, this is the same code as the serial comms, so I figured whatever behavior it did was best to follow since code that uses serial comm also uses socket comm.

autosportlabs/racecapture/api/rcpapi.py
ryandoherty: 👍

autosportlabs/racecapture/databus/databus.py
ryandoherty: Not sure, will check.

autosportlabs/comms/commsfactory.py
ryandoherty: Ah, yeah I did this to use wifi on my mac. Will uncomment and remove socket stuff from this before I merge in.

autosportlabs/comms/socket/socketcomm.py
ryandoherty: 👍

autosportlabs/comms/socket/socketconnection.py
ryandoherty: According to python docs, out of range indexes when slicing results in a blank characters and no exceptions thrown, so we're safe.

autosportlabs/comms/socket/socketcomm.py
brentpicasso: ok

autosportlabs/comms/socket/socketconnection.py
brentpicasso: fantastical!

autosportlabs/comms/socket/socketcomm.py
ryandoherty: From what I can tell, the RCP API code will restart connections (BT, USB and WiFi) if they fail or stop responding. Right now I don't know what types of exceptions would occur that would be safe to continue to process data with. So I think we're ok with leaving this behavior in as it's the same as the BT & USB behavior. 

Once I have a WiFi module to do some real testing against I can see what types of errors may occur that would warrant keeping the thread running.

autosportlabs/racecapture/databus/databus.py
ryandoherty: I commented this out so the sample worker wouldn't poll the mock wifi service I wrote. Un-commented.

ryandoherty:
Ok @brentpicasso I have added comments, fixed the things that needed to be fixed and swapped the polling support to not use is_mobile() and instead query the comms object (for now, long term we need the RC firmware API to expose this).

autosportlabs/racecapture/views/configuration/rcp/scriptview.kv
```diff
@@ -13,28 +13,28 @@
             ScrollContainer:
                 size_hint_y: 0.75
                 do_scroll_x:False
-                id: luaScriptScrollView
```
brentpicasso: what's going on here? merge stuff that's out of sync with other things?

brentpicasso:
Looked; your changes look good but it seems there's extra stuff here from previous changes (around script window, etc). Maybe not 100% up to sync with branch?

ryandoherty:
Oops yeah it's because i merged master into this branch but not the branch
this PR was setup to land in. Will fix.

-Ryan
On May 11, 2016 4:12 PM, "Brent Picasso" <notifications@github.com> wrote:

Looked; your changes look good but it seems there's extra stuff here from
previous changes (around script window, etc). Maybe not 100% up to sync
with branch?

—
You are receiving this because you authored the thread.
Reply to this email directly or view it on GitHub
<https://github.com/autosportlabs/RaceCapture_App/pull/977#issuecomment-218617278>

ryandoherty:
Closing, will merge manually into master.

PR #995, opened 2016-05-24
Choose connection type

ryandoherty:


brentpicasso:
assuming this is the same code as before, so.. :shaved_ice: !

PR #997, opened 2016-05-26
Change socket connection to read only one line/message at a time when reading data

ryandoherty:
It turns out that reading data from a socket isn't the same as reading data from USB. This PR makes a slight change to the socket connection code. Now when the socket connection is asked to read, it will only read one line, then return that data, leaving any extra data behind for use later.

Simple fix, seems to make socket comms stuff work well.

### DIFF
--- a/autosportlabs/comms/socket/socketconnection.py
+++ b/autosportlabs/comms/socket/socketconnection.py
@@ -31,6 +31,7 @@ class SocketConnection(object):
 
     def __init__(self):
         self.socket = None
+        self._data = ''
 
     def get_available_devices(self):
         """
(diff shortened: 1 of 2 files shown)


autosportlabs/comms/socket/socketconnection.py
```diff
@@ -104,14 +104,16 @@ def read(self, keep_reading):
...
 
-                msg += data
+                self._data += data
+
+                if '\r\n' in self._data:
+                    lines = self._data.split('\r\n', 1)
+                    msg = lines[0]
+                    self._data = lines[1]
+
+                    Logger.info("SocketConnection: returning data {}".format(msg))
```
brentpicasso: should be debug maybe? this would generate lots of messages

brentpicasso:
only one note; address and then :shipit:

autosportlabs/comms/socket/socketconnection.py
ryandoherty: 👍

PR #998, opened 2016-05-26
Fix bug where telemetry failed to connect when preference was changed after RC connected

ryandoherty:


main.py
```diff
@@ -92,6 +92,8 @@ class RaceCaptureApp(App):
 
     _status_pump = StatusPump()
 
+    _rc_connected = False
```
brentpicasso: hm, is there no other good way to know if we're connected or not?

main.py
```diff
@@ -404,6 +406,7 @@ def rc_detect_win(self, version):
             self.showStatus("{} v{}.{}.{}".format(version.friendlyName, version.major, version.minor, version.bugfix), False)
             self._data_bus_pump.start(self._databus, self._rc_api, self._rc_api.comms.supports_streaming)
             self._status_pump.start(self._rc_api)
+            self._rc_connected = True
```
brentpicasso: could we put 

self._telemetry_connection.data_connected = True

here instead?

main.py
ryandoherty: We have a few other variables tracking some information we get from RCP, but nothing explicit. Best we have is if the config is loaded, but that sticks around after a disconnect. I've dug around, can't seem to find anything that works, but if there is I'm open to using it.

main.py
ryandoherty: Maybe, will test.

main.py
```diff
@@ -490,6 +491,7 @@ def _on_preferences_change(self, menu, config, section, key, value):
                 if self.rc_config.connectivityConfig.cellConfig.cellEnabled:
                     alertPopup('Telemetry error', "Turn off RaceCapture's telemetry module for app to stream telemetry.")
                 self._telemetry_connection.telemetry_enabled = True
+                self._telemetry_connection.data_connected = True
```
brentpicasso: Hm, does this do anything weird or is it harmless?

ryandoherty: It works fine from my testing, although it does seem wrong to do it on prefs change because RCP may not have connected yet.

ryandoherty: This is why I added a `_rc_connected` variable to track if the app is connected, then it's pretty explicit.

brentpicasso: right.  So is this line needed at all? since we'll be setting it above once we've connected?

ryandoherty: Nope, it's not, fixing.

brentpicasso:
:shipit:

PR #1005, opened 2016-06-06
Fix broken tests

ryandoherty:
Since tests weren't running automatically and people didn't run them before committing, they broke. This fixes them. Mostly small fixes but a second set of eyes would be good.


stieg:
:shipit:

PR #1006, opened 2016-06-07
Datastore security and fixes

### LINKED ISSUE #901: Spaces in channel names break analysis view
Hit this issue while investigating a customer issue reported on the forums: https://autosportlabs.org/viewtopic.php?t=4717&start=0

```
[INFO              ] [DataStore   ] Created session with ID: 1
 Exception in thread Thread-8:
 Traceback (most recent call last):
   File "/usr/lib64/python2.7/threading.py", line 813, in __bootstrap_inner
     self.run()
   File "/usr/lib64/python2.7/threading.py", line 766, in run
     self.__target(*self.__args, **self.__kwargs)
   File "/home/stieg/Devel/asl/rc_app/autosportlabs/racecapture/views/analysis/addstreamview.py", line 183, in _loader_thread
     session_id = self.datastore.import_datalog(logpath, session_name, session_notes, self._update_progress)
   File "/home/stieg/Devel/asl/rc_app/autosportlabs/racecapture/datastore/datastore.py", line 23, in wrap
     ret = f(*args)
   File "/home/stieg/Devel/asl/rc_app/autosportlabs/racecapture/datastore/datastore.py", line 797, in import_datalog
     self._handle_data(dl, headers, session_id, warnings, progress_cb)
   File "/home/stieg/Devel/asl/rc_app/autosportlabs/racecapture/datastore/datastore.py", line 681, in _handle_data
     cur.executemany(datapoint_sql, datapoint_iter(newdata_gen, starting_datalog_id))
 OperationalError: near "Shock": syntax error
```

ryandoherty:
This PR does 2 major things:

1. Fixes #901 , spaces in channel names break importing data to the app
2. Scrubs channel names used for SQL column names and changes all queries to use parameterized values for additional SQL security.

### DIFF
--- a/autosportlabs/racecapture/datastore/datastore.py
+++ b/autosportlabs/racecapture/datastore/datastore.py
@@ -230,6 +230,7 @@ class Filter(object):
     @add_combop
     def group(self, filterchain):
         self._cmd_seq += '({})'.format(str(filterchain).strip())
+        self.params = self.params + filterchain.params
         return self
 
 
@@ -730,12 +731,12 @@ class DataStore(object):
         
     def get_channel_average(self, channel, sessions=None):
         c = self._conn.cursor()
-        params = [channel]
+        params = []
 
         if sessions is not None:
             params = params + sessions
 
-        base_sql = "SELECT AVG(?) from datapoint " + self._session_select_clause(sessions)
+        base_sql = "SELECT AVG({}) from datapoint ".format(self._scrub_sql_value(channel)) + self._session_select_clause(sessions)
         c.execute(base_sql, params)
         res = c.fetchone()
         average = None if res == None else res[0]
--- a/autosportlabs/racecapture/datastore/datastore.py
+++ b/autosportlabs/racecapture/datastore/datastore.py
@@ -903,7 +903,7 @@ class DataStore(object):
         #Put together the smoothing map
         for ch in channels:
             sr = self.get_channel_smoothing(ch)
-            smoothing_map[ch] = sr
+            smoothing_map[self._scrub_sql_value(ch)] = sr
 
         #add the session_id to the smoothing map with a smoothing rate
         #of 0
--- a/autosportlabs/racecapture/datastore/datastore.py
+++ b/autosportlabs/racecapture/datastore/datastore.py
@@ -151,6 +151,7 @@ class Filter(object):
         self._cmd_seq = ''
         self._comb_op = 'AND '
         self._channels = []
+        self.params = []
 
     @property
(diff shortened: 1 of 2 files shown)


autosportlabs/racecapture/datastore/datastore.py
```diff
@@ -347,13 +355,15 @@ def _create_tables(self):
 
         self._conn.commit()
 
+    def _scrub_sql_value(self,value):
+        return ''.join(chr for chr in value if chr.isalnum())
```
Jeff-Ciesielski: Two things: 
* `chr` is a reserved keyword in python, best use something else
* passing this without an additional set of parens seems like a bug that might only work via coincidence.

autosportlabs/racecapture/datastore/datastore.py
```diff
@@ -347,13 +355,15 @@ def _create_tables(self):
...
     
     def _add_extra_indexes(self, channels):
         extra_indexes = []
         for c in channels:
             c = str(c)
             if c in self.EXTRA_INDEX_CHANNELS:
-                extra_indexes.append(c)
+                extra_indexes.append(self._scrub_sql_value(c))
```
Jeff-Ciesielski: Looking at it in context, I think scrub_sql_value probably stands on its own and doesn't need to exist as a class member.

autosportlabs/racecapture/datastore/datastore.py
ryandoherty: Cool, will change `chr`. I'm not sure what you mean about the additional set of parens, from what I can find online about list comprehensions, it looks normal? But I don't have a ton of Python experience, so I'm probably missing something.

autosportlabs/racecapture/datastore/datastore.py
ryandoherty: 👍

autosportlabs/racecapture/datastore/datastore.py
Jeff-Ciesielski: The syntax is correct, but typically you'd surround the list comp with either a set of brackets (creating a new list) or a set of parens (creating a generator).  So:

`''.join((x for x in value if x.isalnum()))` or `''.join([x for x in value if x.isalnum()])`.  I'm just a little confused as to why the interpreter takes it as-is.

ryandoherty:
Ok, @Jeff-Ciesielski, fixes made, lmk if there's anything else, thanks!

ryandoherty:
Merged into r/1.5.4

PR #1015, opened 2016-06-11
Detect if RC firmware didn't find a track and add it automagically to the config so it JUST WORKS

ryandoherty:
Shazam!

* Our users often forget to set their track configs
* Our competition automatically includes the most common tracks in their systems,
* Users shouldn't have to remember to configure their system when it's 2016 and we have the tools and capabilities to handle doing basic searching over a few hundred track locations.

This PR does one thing and does it well. It watches RC hardware when connected and if after all preconditions are met (GPS lock, RC connected, no track detected, tracks loaded in app, etc) it will try to find a nearby track and push it to the connected RC device so lap timing works.

Thing that need help/review:

- How the heck do we sync configuration changes between multiple components? The TrackHelper object is modifying the RC Config tracks db, so if you are viewing/editing your RC configuration and the TrackHelper finds your track, the config view needs to update. My current method is hacky. Ideally I want events to fire *from* the RC config so when it changes, everything is notified. This is problematic if something is listening for changes and making changes at the same time.

### DIFF
--- a/autosportlabs/racecapture/tracks/trackhelper.py
+++ b/autosportlabs/racecapture/tracks/trackhelper.py
@@ -1,0 +1,165 @@
+from datetime import datetime
+from autosportlabs.racecapture.geo.geopoint import GeoPoint
+from kivy.logger import Logger
+from kivy.event import EventDispatcher
+from autosportlabs.racecapture.config.rcpconfig import *
+
+
+class TrackHelper(EventDispatcher):
+    """
+    This class' responsibility is to make sure RC detects a track when it runs so
+    users get lap timing when they are on track
+    """
+
+    def __init__(self, track_manager, rc_api, status_pump, rc_config, **kwargs):
+        """
+        Init!
+        :param track_manager: TrackManager instance
+        :param rc_api: RCAPI instance
+        :param status_pump: StatusPump instance
+        :param rc_config: RCConfig instance
+        :param kwargs: kwargs
+        :return: None
+        """
+        self.register_event_type('on_track_detected')
+        super(TrackHelper, self).__init__(**kwargs)
+
+        self.track_manager = track_manager
+        self.rc_api = rc_api
+        self.status_pump = status_pump
+        self.rc_config = rc_config
+
+        # Number of seconds to wait until we attempt to figure out what track we're at,
+        # allows RC hardware some time to detect a track
+        self._wait_period = 4
+
+        #When RC was ready, we wait _wait_period after this to ensure RC had time to search for a track
+        self._ready_start_time = None
+
+        #Status hash from StatusPump
+        self._status = None
+
+        #Has the TrackManager loaded all tracks?
+        self._tracks_loaded = False
+
+        #Did RC detect a track?
+        self._track_detected = False
+
+        self.status_pump.add_listener(self.on_status)
+
+        if len(self.track_manager.tracks) == 0:
+            self.track_manager.load_tracks(success_cb=self._on_tracks_loaded, fail_cb=self._on_tracks_load_fail)
+            Logger.info("TrackHelper: Loading tracks")
+        else:
+            self._tracks_loaded = True
+
+        Logger.debug("TrackHelper: initialized")
+
+    def on_track_detected(self):
+        """
+        Dispatched when track found in track db
+        :return:
+        """
+        pass
+
+    def on_status(self, status):
+        """
+        StatusPump status listener
+        :param status: Status hash
+        :return: None
+        """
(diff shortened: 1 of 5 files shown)


ryandoherty:
Oops, set this against master, will merge into 1.6.0 when reviewed.

autosportlabs/racecapture/tracks/trackhelper.py
```diff
@@ -0,0 +1,165 @@
...
+            return False
+
+        return True
+
+    def _check_track_detection(self):
+        """
+        If everything is ready, attempt to find a track in the app's local track db that is near
+        where RC is now. Uses the track manager to find a track, if found adds it to RC Config's
+        track db and fires the 'on_track_detected' event
+        :return:
```
brentpicasso: Instead of adding this to the list of auto-detected track DB, I think we should write it to the custom / user-defined track config. 

This is especially important if we pull the track DB mechanism from RCT and 100% rely on the app to detect and set the track. IMO, this should be done anyway, since RCT is designed to rely on the app for operation, whereas we still need it on RCP as it's designed to operate autonomously.

The side benefit is faster config update vs the more complicated process of re-writing the track DB.

ryandoherty: So how does the firmware use the user defined track setting? If we set it,
then the user goes to a different track and *does* have that track in its
track database, will it do the right thing and use the correct start finish?

-Ryan
On Jun 11, 2016 10:51 AM, "Brent Picasso" <notifications@github.com> wrote:

In autosportlabs/racecapture/tracks/trackhelper.py
<https://github.com/autosportlabs/RaceCapture_App/pull/1015#discussion_r66709350>
:

> +
> +        if diff.total_seconds() < self._wait_period:
> +            Logger.debug("TrackHelper: waiting for RC warmup")
> +            return False
> +
> +        if not self._tracks_loaded or not self.rc_config.trackConfig:
> +            return False
> +
> +        return True
> +
> +    def _check_track_detection(self):
> +        """
> +        If everything is ready, attempt to find a track in the app's local track db that is near
> +        where RC is now. Uses the track manager to find a track, if found adds it to RC Config's
> +        track db and fires the 'on_track_detected' event
> +        :return:

Instead of adding this to the list of auto-detected track DB, I think we
should write it to the custom / user-defined track config.

This is especially important if we pull the track DB mechanism from RCT and
100% rely on the app to detect and set the track. IMO, this should be done
anyway, since RCT is designed to rely on the app for operation, whereas we
still need it on RCP as it's designed to operate autonomously.

The side benefit is faster config update vs the more complicated process of
re-writing the track DB.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/autosportlabs/RaceCapture_App/pull/1015/files/abf62a9157a51dbd95eb6a9ceb8c4b2b39b094f6#r66709350>,
or mute the thread
<https://github.com/notifications/unsubscribe/AACyPEdrpXeuVgUxpaqTIJjPLxpNC5Akks5qKvW5gaJpZM4Iza4D>
.

brentpicasso: I was thinking about this too, and we may need to adjust the behavior of the firmware to make it work. 

RCT:  simply use what's in the fixed config, if values are set (non zero)
RCP: if not found in track DB, use what's in the fixed config, if values are set 

Might need to meet about this to fully flesh out the edge cases. I think the bulk of what you have here is on the right track. (get it!!?! :D )

ryandoherty: That behavior will cause some problems and user confusion for a few reasons (mostly for RCP, some for RCT depending on if we actually do remove the track db from its firmware):

- Currently the app has a binary choice between autodetect and manual track configuration. This is a very simple and easy to understand distinction for users. This new behavior adds an additional rule to understand that will definitely confuse them. "I turned on autodetect and it used my previously set manual start/finish."
- RCP would use the manual start/finish when autodetect is on but no track was detected, so the app would show 'User defined' even though autodetect is on, causing confusion. 

The big benefit of the behavior for this PR is is requires no firmware changes. 

And regarding the related track detection for RCT, even with a limited # of tracks that RCT could store (which could be improved via removing things like sector information and other metadata), odds are very low that anyone would visit more than 10 or 20 different tracks. 

This PRs behavior is the simplest behavior for users to understand and simplest in terms of code complexity. Both are big advantages imho.

ryandoherty:
Closing this, we're using other mechanisms to achieve this (setActiveTrack and some other app behaviors)

PR #1021, opened 2016-06-22
Adjust config screen to only show configs applicable to connected hardware

ryandoherty:
Holy cow this ended up being bigger than originally imagined. Much was changed, much was improved.

tldr; When a RC device connects, the config view will look at the `getCapabilities` command response to see what hardware the device has, and only display the config items that are applicable.

The diff is large, but conceptually 5 things have changed:

1. RCP API class checks capabilities before querying for configuration so we don't get timeouts when querying for configs that don't exist
2. RCP Config class now has a Capabilities class that stores all the capabilities information and has convenience wrappers like `has_bluetooth` for consumers of the class to use.
3. The Config view now checks the Capabilities object to determine what menu items to show and what wireless configuration sections to display. 
4. The Wireless Config views have been refactored into distinct classes and views for simplicity. They are redrawn when the configuration changes to reduce code that manages state.
5. Lots of UI cleanup was done when refactoring the Wireless Config views (BT, WiFi, Cell). The Wireless config UI is a lot cleaner looking, this affected other layouts too. 

I'd say this is about 95% ready, there are probably hidden gremlins that I can't see because I've been staring at this code for weeks.


autosportlabs/racecapture/api/rcpapi.py
```diff
@@ -169,7 +170,12 @@ def msg_rx_worker(self):
                 if msg:
                     # clean incoming string, and drop illegal characters
                     msg = unicode(msg, errors='ignore')
-                    Logger.debug('RCPAPI: Rx: ' + str(msg))
+
+                    if 's' in msg:
```
brentpicasso: What's the 's' for?

autosportlabs/racecapture/config/rcpconfig.py
```diff
@@ -1016,59 +1016,111 @@ def is_valid(self):
...
 class ChannelCapabilities(object):
-    analog = 8
-    imu = 6
-    gpio = 3
-    timer = 3
-    pwm = 4
-    can = 2
+
+    def __init__(self):
+        self.analog = 7
```
brentpicasso: why did we change this down to 7? MK2 has 7 channels + the internal battery channel

autosportlabs/racecapture/views/configuration/rcp/canconfigview.py
```diff
@@ -17,7 +19,7 @@ class CANBaudRateSettings(SettingsView):
 class CANBaudRateSpinner(SettingsMappedSpinner):
     def __init__(self, **kwargs):    
         super(CANBaudRateSpinner, self).__init__(**kwargs)
-        self.setValueMap({50000: '50K Baud', 100000: '100K Baud', 125000: '125K Baud', 250000:'250K Baud', 500000:'500K Baud', 1000000:'1M Baud'}, 500000)
+        self.setValueMap({50000: '50K Baud', 100000: '100K Baud', 125000: '125K Baud', 250000:'250K Baud', 500000:'500K Baud', 1000000:'1M Baud'}, '500000')
```
brentpicasso: the default value (a the end of line 22) changed to string, intentional?

autosportlabs/racecapture/views/configuration/rcp/configview.py
```diff
@@ -99,8 +103,29 @@ def on_channels_updated(self, runtime_channels):
         self.update_runtime_channels(runtime_channels)
 
     def on_config_updated(self, config):
-        self.rc_config = config
-        self.update_config_views()
+        if config.versionConfig.serial != self._sn:
+            # New device, redraw
+            # Our config object is the same object with new values, so we need to copy our value
+            self._sn = copy(config.versionConfig.serial)
```
brentpicasso: is the serial a regular string or a dict? if a string, they're immutable in python so you shouldn't need to do the copy.

brentpicasso:
I see inconsistent font sizes and inconsistent fonts on the wireless config screen:

image

brentpicasso:
When RC app starts up with no device connected, it shows all options as defaulted by MK2.  What about showing no menu items, except for firmware upgrade until a config is read from device or loaded from file?

autosportlabs/racecapture/views/configuration/rcp/wireless/wificonfigview.py
```diff
@@ -0,0 +1,146 @@
+import kivy
+kivy.require('1.9.1')
+from kivy.app import Builder
+from kivy.uix.gridlayout import GridLayout
+from kivy.logger import Logger
+import os, json
+from settingsview import SettingsView, SettingsSwitch, SettingsButton, SettingsMappedSpinner
+from autosportlabs.uix.bettertextinput import BetterTextInput
+
+WIFI_CONFIG_VIEW = 'autosportlabs/racecapture/views/configuration/rcp/wireless/wificonfigview.kv'
```
brentpicasso: We should include the .kv inline with the file, inline with the new standard / style

autosportlabs/racecapture/api/rcpapi.py
ryandoherty: That's a sample, when I was using debug log level you see samples flying by at 50hz and miss anything important.

autosportlabs/racecapture/config/rcpconfig.py
ryandoherty: If internal battery channel is there by default, should we even list it? I'll check what mk2 lists as the # of analog channels.

autosportlabs/racecapture/views/configuration/rcp/canconfigview.py
ryandoherty: Yeah, I was getting an error stating that a spinner's value had to be text, so I changed it. Not sure why it popped up now.

autosportlabs/racecapture/views/configuration/rcp/configview.py
ryandoherty: String. The config object is never re-instantiated, so the serial property is being overwritten when a config is read. I had to do this to keep ahold of the previous value.

ryandoherty:
Will dig into the font size issue, that is weird.

I had thought about what to show if nothing is connected. I'm thinking showing everything by default is best so people who haven't bought a RC can play with the app and if you need to load a config to view it, you can (I don't think the capabilities info is saved to a config file).

autosportlabs/racecapture/views/configuration/rcp/wireless/wificonfigview.py
```diff
@@ -0,0 +1,146 @@
...
+        Logger.debug("WifiConfigView: got wifi enabled change")
+        if self.wifi_config:
+            self.wifi_config.active = value
+            self.wifi_config.stale = True
+            self.dispatch('on_modified')
+
+    def build_ap_channel_spinner(self, spinner):
+        channel_map = {}
+
+        for i in range(1, 12, 1):
```
brentpicasso: I wonder, for consistency, if we should externalize these values like we do with ap_encryption.json, even though it's a numeric array. thoughts?

autosportlabs/racecapture/views/configuration/rcp/wirelessconfigview.py
```diff
@@ -1,278 +1,68 @@
...
+        self.base_dir = base_dir
+        self.rcp_capabilities = capabilities
+        self.rcp_config = config
+        self._views = []
+
+        self._render()
+        self._attach_event_handlers()
+
+    def _render(self):
+        if not self.rcp_capabilities or (self.rcp_capabilities and self.rcp_capabilities.has_bluetooth):
```
brentpicasso: so we show these sections by default if capabilities isn't loaded?

autosportlabs/racecapture/views/configuration/rcp/wireless/wificonfigview.py
ryandoherty: 👍

autosportlabs/racecapture/config/rcpconfig.py
brentpicasso: it should show 1 channel. We include the battery channel so it can be calibrated by the end user if necessary (it's come up in the past)

autosportlabs/racecapture/views/configuration/rcp/wireless/wificonfigview.py
ryandoherty: Eh, seems silly for something this simple.

autosportlabs/racecapture/views/configuration/rcp/canconfigview.py
brentpicasso: ok

autosportlabs/racecapture/views/configuration/rcp/wirelessconfigview.py
ryandoherty: Hmm, this may be super early code when I built this, I think the check is unnecessary. We should show it if no device is connected (default capabilities are used)

autosportlabs/racecapture/views/configuration/rcp/configview.py
brentpicasso: Gotcha.

autosportlabs/racecapture/config/rcpconfig.py
ryandoherty: 👍

brentpicasso:
Actually, capabilities are saved to a config file, so this idea should work.  IMO we shouldn't confuse people with options they can't use.

ryandoherty:
Actually capabilities aren't saved to a config file and they shouldn't be so you can apply a configuration from RCT to RCP and vice versa (with the configs that are applicable to each). 

I'm not sure what you are referring to regarding show users options they can't use. Once a device is plugged in the UI will update to the appropriate config.

brentpicasso:
Looking through the screens again.  I'm seeing the buttons for CAN baud rates with no spacing between them now:

image

brentpicasso:
If I read the config from RCT, and then read it again, I see this in the wireless view: 
image

brentpicasso:
Also, I noticed Write button is activated once I enter the wireless config view. Seems that an event is getting fired that makes it's way up to enable the button.

brentpicasso:
STR:  

* run app
* from home screen allow config to load (this seems to be the important step)
* navigate to config view
* select wireless view
* click read button, click yes to confirm reading
* observe menu reloading / double adding of widget panels

brentpicasso:
When trying to save a config I get this error.

When saving a config on 1.5.3 and MK2, I don't get this error, and the capabilities block is saved to the file.  I think it's important that capabilities remain saved to the config file so the UI can be rendered appropriately when *just* a config file is loaded.


image

ryandoherty:
If we hide configs when we load a saved config file, we might end up hiding configuration options that are applicable to the connected hardware. So we need to adhere to the connected hardware's capabilities, not the config file's.

brentpicasso:
That's reasonable - if the device is connected then that should override the view settings. 

However, IMO we need to handle the case when the app is running, no device is connected, and we load a config file, so the view accurately represents what was loaded.

brentpicasso:
Issue observed.

If you click read config without changing hardware, it will redraw menu and wipe/re-create screens. this only happens once, under certain conditions: 

Start app with RC or RCP connected
allow config to be fully read
navigate to config mode
click a config view, like, scripting
after screen loads, click read button
observe menu re-loading and the user now sees the RaceTrack config screen.

brentpicasso:
Also, we should show the analog page so they can at least make adjustments to the battery voltage, if we want that to be adjustable.  However, we can choose to make it fixed at the current 1Hz, to simplify the UI for now.

brentpicasso:
Need to consistently use the ASL fonts for the wifi config screen:

image

brentpicasso:
One more issue: I cannot set the telemetry device ID

brentpicasso:
minor UI: Should have margin on the right of the CAN baud rate button:

image

ryandoherty:
Ok, *think* it's nearly g2g. You can now set telemetry device id, CAN settings have padding and analog channels are shown with appropriate # of channels.

brentpicasso:
Very close. just need to tweak the size of the analog sensor screen: 

image

brentpicasso:
once this is done I think we can merge this thing.

brentpicasso:
By the power of greyskull, :shipit:

PR #1024, opened 2016-06-27
Analysis by time and other improvements

### LINKED ISSUE #782: Time based graphing
There are some cases where we don't have a distance channel, or want to analyze logs where distance is not incrementing. 

Ideas:
* Allow the charts to show time based data
* Split data into manageable chunks over long periods of time, in order to handle large data logs. Allow paging between chunks of data, similar to laps.
* For long data files, show a lower resolution higher summary view with a sliding window that will show what page you're currently on

brentpicasso:
* Provide analysis by time and distance in chart
* Switchable via tool-bar and options menu for chart
* Extended options menu for chart to include extra options screens (we'll be adding more for other things)
* Clarified toggle buttons on options screen
* Streamlined channel selection for chart widgets, removing extra steps to select new channels
* Provide a time or distance marker that follows the current drag position
* Extracted features and general refactoring - e.g. creating a central options screen for different widgets
* Fixed bug where a channel was selected where it's non-existant in a session

Resolves #782


autosportlabs/racecapture/views/analysis/channelvaluesview.py
```diff
@@ -72,7 +72,11 @@ def value(self):
 
     @value.setter
     def value(self, value):
-        self.value_view.value = float(value)
+        try:
+            value = float(value)
+            self.value_view.value = value
+        except:
```
ryandoherty: I'm really not a big fan of generic exception catching like this, you can end up catching things you didn't expect and weird, unexpected behavior can occur (app could continue running in an unknown or bad state, causing other errors or data corruption, etc). Also makes debugging really hard because there is no logging. Or if you ask a customer for logs, their may not be anything useful because you caught everything and didn't log it.

If you know what may go wrong in your `try` block, catch that and log it (as debug or trace if it's something super noisy or not important). 

Related: http://www.martinfowler.com/ieeeSoftware/failFast.pdf

brentpicasso: I'm following the Python convention of ask for forgiveness vs permission. In a lot of cases, it's actually faster code.
http://stackoverflow.com/questions/12265451/ask-forgiveness-not-permission-explain

However, what I should be doing is catching the specific float conversion exception, b/c that's really what my intent was.  I can update to account for that.

autosportlabs/racecapture/views/analysis/customizechannelsview.py
```diff
@@ -23,169 +23,159 @@
...
+    Builder.load_string(KV_FILE)
+
     def __init__(self, **kwargs):
-        super(CurrentChannelsView, self).__init__(**kwargs)
-        self.register_event_type('on_confirm_customize')
-        self.register_event_type('on_add_channels')
+        super(CustomizeChannelsView, self).__init__(**kwargs)
+        self._channel_widgets = {}
+        self.register_event_type('on_channels_customized')
+        self.datastore = kwargs.get('datastore')
```
ryandoherty: If a keyword arg is required for this to run, it should be a named arg so it's obvious and tools can introspect the code to know what this class requires. http://docs.python-guide.org/en/latest/writing/style/#function-arguments

brentpicasso: Copied and pasted from older code during refactoring.  Can update :+1:

brentpicasso: fixed.

ryandoherty: Thanks! 👍

autosportlabs/racecapture/views/analysis/linechart.py
```diff
@@ -15,55 +15,71 @@
...
     def __str__(self):
-        return "{}_{}".format(str(self.sourceref), self.channel) 
+        return "{}_{}".format(str(self.sourceref), self.channel)
+
+class LineChartMode(object):
+    '''
+    Describes the supported display modes for the chart
+    '''
+    time = 1
+    distance = 2
```
ryandoherty: Should these be an enum or at least ALL CAPS to denote they are special?

autosportlabs/racecapture/views/analysis/linechart.py
```diff
@@ -15,55 +15,71 @@
...
     '''
     Displays a line chart capable of showing multiple channels from multiple laps
     '''
     color_sequence = ObjectProperty(None)
-    _channel_plots = {}
-    ZOOM_SCALING = 0.02
+    ZOOM_SCALING = 0.01
     TOUCH_ZOOM_SCALING = 0.000001
-    
+    MAX_SAMPLES_TO_DISPALY = 1000
```
ryandoherty: Typo.

autosportlabs/racecapture/views/analysis/linechart.py
```diff
@@ -151,34 +196,47 @@ def dispatch_marker(self, x, y):
...
         self.ids.chart.marker_x = data_index
+
+        # Label marker that follows marker position
+        marker_x = x
+        if self.line_chart_mode == LineChartMode.time:
+            marker_x -= self.MARKER_LABEL_WIDTH_TIME if self.width - marker_x < self.MARKER_LABEL_WIDTH_TIME else 0
+            marker_value = format_laptime(data_index * 0.000016666)
+        else:
+            marker_x -= self.MARKER_LABEL_WIDTH if self.width - marker_x < self.MARKER_LABEL_WIDTH else 0
+            marker_value = '{:.2f}'.format(data_index)
```
ryandoherty: 0.000016666 looks like a magic number. What is it?

I do a lot of these if/else things in Podium's charting stuff for time or distance charting. Wonder if we can extract this out into class vars that are changed when the user changes the graphing mode? Then you remove all if/else statements to handle distance or time values.

You'd have something like this:

`marker_x -= self.marker_label_width if self.width - marker_x < self.marker_label_width else 0 
 marker_value = format_laptime(data_index * self.conversion_value)
`

Then in `on_toggle_chart_mode` you would change  `self.marker_label_width` and `self.conversion_value` to the time or distance values. Thoughts?

autosportlabs/racecapture/views/analysis/linechart.py
```diff
@@ -187,103 +245,328 @@ def on_touch_move(self, touch):
...
+
+                channel_plot.chart_x_index = time_index
+                plot.ymin = channel_data_values.min
+                plot.ymax = channel_data_values.max
+                plot.points = points
+                self._channel_plots[str(channel_plot)] = channel_plot
+
+                # sync max chart x dimension
+                self._update_max_chart_x()
+        finally:
```
ryandoherty: Try but no catch, what can go wrong here?

ryandoherty:
I'm having trouble wrapping my head around how the new customize and options screens are used. Some code examples as comments would help so future devs can re-use them easily.

autosportlabs/racecapture/views/analysis/linechart.py
brentpicasso: By design.  anything can go wrong, and the crash handler will handle it, but we must decrement the progress spinner or it will leave the spinner on the screen forever.

brentpicasso:
Good point, will document those optionscreen classes.

autosportlabs/racecapture/views/analysis/linechart.py
brentpicasso: Can do.

autosportlabs/racecapture/views/analysis/linechart.py
ryandoherty: Ok.

ryandoherty:
Found a weird bug, a number is following my mouse cursor:
http://i.imgur.com/L8YQmuZ.png
Happens when I load the analysis view, even after restarting the app.

brentpicasso:
Select a lap and you'll see what the intention is behind that.  Looks like we need to blank that out if no laps are selected.

ryandoherty:
It also shouldn't show up if the mouse isn't over the graph. I'm not sure what the value is, usually when you hover over a graph with a mouse, it shows the graph values, but this looks like the x index?

brentpicasso:
it's the distance or time value.  Watch it when you switch between distance or time modes.

ryandoherty:
My point is it's not really useful or obvious as to what it is. People care about the Y values (throttle, speed, etc). Distance or time really isn't important enough to show next to the mouse cursor.

ryandoherty:
Confusing behavior I just understood, the channel list re-orders itself when you unselect an item. Also, you can select an item by tapping anywhere on the list item, but to remove you have to tap the garbage can. A more common behavior is to to tap again to unselect, no special area to remove needed.

brentpicasso:
I purposely put the already selected items at the top of the list, always - so you don't have to search the list of channels to unselect them. 

Can make the whole label apply to the de-selection.

ryandoherty:
Yeah, I understood what was going on, it was confusing because the item would look like it disappeared when I unselected it. If we want to make the selected items be easier to find, having a separate 'selected' section or divider between the selected and unselected items would help visually delineate between the groups.

Issue #1020, opened 2016-06-20
Remove Tracks top level menu item and homepage icon


ryandoherty:
Some users are getting confused by the top level Tracks icon, which manages just the tracks saved to the app. They think they are managing RCP's configuration, which they are not. We should remove it to prevent this confusion. Users can still update tracks from the Configuration screen.

PR #1030, opened 2016-06-28
Remove tracks menu item

ryandoherty:
To remove any confusion as to how a user adds tracks to RCP, this PR removes the homepage and menu items that would take you to the tracks manager. Now, you only can update and select tracks in the configuration screen.

Also changes 'Dashboard' to 'Race' on the homepage view, a step in the direction of a 'race' mode that will allow the app to know that the user is heading on to the track and to do a final check of all requirements.


brentpicasso:
Resolves #1020

brentpicasso:
:shipit:

PR #1031, opened 2016-06-29
High performance tachometer using new widget design + bug fixes + refactorings

brentpicasso:
* Improves performance of tachometer rendering by using the openGL reveal technique seen in the round gauges
* Alert  / warning range coloring now works 
* Channel can be selected or de-activated
* Other refactorings to align with coding standards and style.


brentpicasso:
re-opening against the correct branch

PR #1032, opened 2016-06-29
High performance tachometer using new widget design + bug fixes + refactorings

brentpicasso:
Opening this against 1.5.4, assuming we'll release as 1.6.0 when we're done with issues.

----------------------

* Improves performance of tachometer rendering by using the openGL reveal technique seen in the round gauges
* Alert  / warning range coloring now works 
* Channel can be selected or de-activated
* Other refactorings to align with coding standards and style.

Resolves #1031


brentpicasso:
video demo: 

https://www.youtube.com/watch?v=m_xnXZBWGNw&feature=youtu.be

autosportlabs/racecapture/views/dashboard/widgets/tachometer.py
```diff
@@ -1,31 +1,177 @@
...
+            PushMatrix()
+            Color(1, 1, 1, 1)
+            self.shadow_translate = Translate(x, y, 0)
+            self.shadow_scale = Scale(x=scale_x, y=scale_y)
+            Rectangle(source='resource/gauge/horizontal_bar_gauge_shadow.png', pos=self.pos, size=(self.gauge_width, self.gauge_height))
+            PopMatrix()
+
+        self.bind(pos=self.update_all, size=self.update_all)
+
+    def update_all(self, *args):
```
ryandoherty: What's the *args used for? Is this function called with different arguments? I don't see any args used in this function.

autosportlabs/racecapture/views/dashboard/widgets/tachometer.py
```diff
@@ -1,31 +1,177 @@
...
+            railed_value = value
+            if railed_value > max_value:
+                railed_value = max_value
+            if railed_value < min_value:
+                railed_value = min_value
+
+            value_range = max_value - min_value
+            offset = railed_value - min_value
+            self.ids.gauge.value = offset * 100 / value_range
+        except Exception as e:
```
ryandoherty: Why are we catching all Exceptions here?

autosportlabs/racecapture/views/dashboard/widgets/tachometer.py
brentpicasso: they're values passed in by the event handler, and we ignore them.

autosportlabs/racecapture/views/dashboard/widgets/tachometer.py
brentpicasso: explained by comment on line 173

autosportlabs/racecapture/views/dashboard/widgets/tachometer.py
ryandoherty: 👍

autosportlabs/racecapture/views/dashboard/widgets/tachometer.py
ryandoherty: So we wouldn't see bugs in Sentry with this exception handler. I think we need a way for exceptions to make it to Sentry so we can see bugs in the wild.

brentpicasso: We can drop it - the sentry errors will get throttled at some point up the chain. Looking to see if the python raven client does internal throttling. 

Also, we should probably log an issue to sweep through and remove exception handlers like this throughout the rest of the code.

ryandoherty: 👍

ryandoherty: I agree having some way to throttle or block repeated exceptions is a good idea, not sure how to handle it though.

ryandoherty:
Overall this looks great and is a huge improvement! :shipit:

PR #1028, opened 2016-06-28
Add left/right keybindings to dashboard view.

ddimensia:
I added the ability to capture the 'left'/'right' keyboard keypresses on the dashboard view so that it could mimic pressing the left/right nav buttons respectively.

PR #1034, opened 2016-06-30
Add left/right keybindings to dashboard view.

ddimensia:
Refer to PR #1028 for previous comments.


ryandoherty:
Hi ddimensia, thanks for staying on top of things and moving the PR to 1.6.0, I need to re-jigger the branches, 1.5.4 is becoming 1.6.0, 1.6.0 => 1.7.0. I'm not sure what will happen to your PR, you may have to recreate it again. Thanks for your patience!

ddimensia:
Just let me know when you have the branches setup the way you want them and I'll make the appropriate PR changes.  The branching strategy I'm used to is to have development on master, and all X.Y.0 releases come from there (e.g. 1.6.0) and at release time create a branch for any patches to that release (e.g. 1.6.x).  Of course major features who's development may span releases should be done in a feature branch and then merged to master when they're ready.

ryandoherty:
Ok, 1.6.0 branch has been remade, can you open a PR against it? Thank you!

The branching strategy you describe is more common and we've talked about what strategy works for us internally many times. The main reason for our current setup is we use semantic versioning (http://semver.org/) and attempt to group our features into appropriate versions. So we often need a 1.1.0 branch in parallel to a 1.0.1 (for example). That way someone can land a big, new feature into 1.1.0 while we land fixes into 1.0.1.

ddimensia:
We use semver also where I work and create minor releases weekly often times with one or more patch release within the week also.   Basically, we keep master as the development branch for the next release, an X.Y release, no patches.  When we cut a release we also create a patch branch at the same time for any future patch releases.  At that point any patches for a release get merged into the patch branch with the expectation that the same fix is made against master (two PRs).   This allows PRs for the next major/minor release to just be against master and reduces the overhead of merging changes forward from release to release (if you create the next release in advance).  
   
```
--1.5.0--------------1.6.0------------> master
    \                   \_1.6.x
     \ _1.5.x                   
```

PR #1037, opened 2016-06-30
Added 'demo' mode that replays a RCP log file.

ddimensia:
See PR #1035 

Also noticed while using this that the new tach display doesn't scale appropriately based on the min/max of the RPM output.


brentpicasso:
Thanks for this PR. The idea of playback is useful, especially in the context of RCP simulation. 
We have an API simulator on the firmware side that creates a virtual serial port that the app can talk to, but it does not support streaming of data:
https://github.com/autosportlabs/RaceCapture-Pro_firmware/blob/master/test/RCPSim.cpp

This would require the ability to build firmware artifacts using gcc, which can be inconvenient for some.

I looked over your implementation.  What I think would be cleaner would be a simulator that can be launched off to the side as a stand-alone tool, and communicate over the new TCP socket interface we built for WiFi connectivity. This tool could act as a TCP server and the app would connect to it over the known port via loopback address. 

By having it as a stand alone tool, it prevents it from being deeply embedded into the app, and could be expanded easily for other purposes and minimize regression risks in the app.  I think the core of what you wrote could be easily adapted to this use. 

Thoughts?

ddimensia:
I personally like the idea of keeping it embedded and taking it a step further by adding UI elements to enable the functionality and select a file.   This would make it work with non-commandline apps and make it easy to test UI elements in tablets etc.  

Building a TCP server would be relatively easy to do also and would be better for exercising the rest of the communication code as you could also support faking/capturing config calls etc.  In some ways it would have similar limitations to the commandline based setup I have now unless you added UI to allow the user to specific IP/Port (not a bad idea).

ddimensia:
I'm assuming if one was to build a TCP server you'd prefer it in Python?

brentpicasso:
Yeah, it's super easy to create a TCP server in python; recommend creating a tools directory off of the root of the project and put the simulator in there. :+1:

ddimensia:
That hurts my static typing with compile time guarantee sensibilities.  I already started a Kotlin + Vert.x version at https://github.com/ddimensia/rcp-server but will consider rewriting it in Python.

brentpicasso:
I love me some static typic and compile time protection, but I have to say, Python is actually fun to code in. :)

brentpicasso:
Going to close this now, if you add a separate simulator as a developer-type tool please feel free to open a new PR. thanks!

PR #1045, opened 2016-07-05
Bluetooth name and password configuration

ryandoherty:
This PR allows users to modify their BT name and password configuration (with some major caveats).

* Added 'advanced config' button to Wireless config view, pops up modal for editing BT config
* BT config advanced view. Has convenience functions for validation and getting all form values.
* `BetterTextInput` pulled into 1.6.0 to allow for easier restriction of form field element input.
* Improved CenteredBubble placement and arrow. 
* SettingsView help text is now right aligned so it is closed to the label and control, making it easier to read.

Caveats:
* RCP firmware can't read the BT module's configuration, so we have no idea what it is. I added text explaining this, but it may not help much.
* Firmware < 2.9.0 doesn't update the BT module's configuration if given a new config. I added text explaining this because in r/1.7.0 we have methods for showing/hiding configuration view info based on capabilities, but moving it into 1.6.0 would be a massive amount of work. So leaving it for now. This will affect few users I believe (mostly mk1).


autosportlabs/racecapture/config/rcpconfig.py
```diff
@@ -1025,6 +1032,16 @@ def from_json_dict(self, json_dict):
...
 
+        # For select features/capabilities we need to check RCP version because
+        # the capability wasn't added to the API. Not ideal, but let's at least
+        # insulate other code from inspecting the version string
+        if version_config:
+            min_bt_config_version = StrictVersion(self.MIN_BT_CONFIG_VERSION)
+
+            rcp_version = StrictVersion(version_config.version_string())
+
+            self.bluetooth_config = rcp_version >= min_bt_config_version
```
brentpicasso: where does this variable get used? I only see it set here and defined on line 1019.

ryandoherty: I was using it for the BluetoothConfigView to hide the 'Advanced' button, but with the huge WirelessConfigView+Capabilities rewrite in 1.7.0, I held off. Once this is merged into 1.7.0 (it will have some conflicts which I'll handle), this variable will be used there.

brentpicasso: Got it.

Think it makes sense to roll forward the idea of hiding the advanced button if firmware < 2.9.0, since we'll need it anyway? That way, you can remove the firmware version warning message.

ryandoherty: I'm not sure what you mean by 'roll forward'. My plan was for 1.6.0 to leave this as-is, but in 1.7.0, hide the advanced button if the firmware is less than 2.9.0. This is because if I had to show/hide the button, I have to show/hide it when different RCPs connect/disconnect/reconnect, etc. The functionality for this is in 1.7.0, so I can do that in that release. Thoughts?

brentpicasso: Sorry. I meant, do the work now to hide the advanced button if < 2.9.0, to avoid the warning message.

I understand the rationale behind leaving it as is until 1.7.0.   Sounds good to me.

ryandoherty: Yeah, that's what I started to do in 1.6.0, but then realized I would either be writing nearly the same code as in 1.7.0, or have to backport the changes from 1.7.0 into 1.6.0. Since this is a feature that most users won't touch and the majority of our users are on 2.9.0 (or can upgrade), having a message about the firmware restrictions for now seems ok to me.

brentpicasso:
Did some testing. Seems to work well. 

Recommend making size of name limited to 14 characters.  

Tried "a1234123412341" and succeeded, but "a12341234123412" failed to update BT module name

Recommend removing period at end of the following:
Minimum 1 charcter
4 digits required
Firmware version 2.9.0 or greater required

brentpicasso:
also, test with ./emu_phone.sh to check phone form factor. looks like dialog for phone form factor is squished, and Advanced BT configuration may need to be shortened / made more concise.

ryandoherty:
Cool, will update per your recommendations and test on mobile 👍

brentpicasso:
:shipit:

PR #1048, opened 2016-07-08
Fix Bluetooth layout bugs, advanced BT config layout and spinner heights

ryandoherty:
Yeah.... so I shot myself in the foot when I messed with some layouts in 1.6.0 and 1.7.0 at the same time. This fixes problems with the BluetoothConfigView and SettingsView, along with fixing spinner heights in the Wireless settings view.


brentpicasso:
:shipit:

PR #1050, opened 2016-07-12
Fixed issue when exceptions not being logged properly within the api code.

ddimensia:
While testing a TCP based log replay server I came across a few cases where a message from the server would fail to parse but the RaceCapture app didn't produce useful output about the exception.  The reason for adding the repr() call around the msg is to output non-printable characters in a way that it can be determined in the logs what the app saw.

### DIFF
--- a/autosportlabs/racecapture/api/rcpapi.py
+++ b/autosportlabs/racecapture/api/rcpapi.py
@@ -130,8 +130,8 @@ class RcpApi:
         try:
             self.comms.close()
             self.comms.device = None
-        except Exception:
-            Logger.warn('RCPAPI: Message rx worker exception: {} | {}'.format(msg, str(Exception)))
+        except Exception as e:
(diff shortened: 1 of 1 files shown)


brentpicasso:
Hi Gil, thanks for the submission!  Is it allowing you the ability to add an issue to match this PR? If so, please do so to track the resolution.  A quick test and we should be able to merge it.

autosportlabs/racecapture/api/rcpapi.py
```diff
@@ -130,8 +130,8 @@ def shutdown_comms(self):
         try:
             self.comms.close()
             self.comms.device = None
-        except Exception:
-            Logger.warn('RCPAPI: Message rx worker exception: {} | {}'.format(msg, str(Exception)))
+        except Exception as e:
+            Logger.warn('RCPAPI: Message rx worker exception during comms shutdown: {}'.format(str(e)))
```
ddimensia: I think the original reference to msg was a copy/paste error so I updated the log statement accordingly.

PR #1058, opened 2016-07-15
Change Logger.info() to debug() for comms and RCAPI info

ryandoherty:
Currently the comms and RCAPI classes are a bit chatty via Logger.info() calls, which makes it hard to debug and develop because they spam the console. This changes their calls debug(). Most of the info() calls were the same information as shown in the status bar in the app, so not much is lost during normal development.


brentpicasso:
:shipit:

PR #1063, opened 2016-07-16
Make LapDelta filter not add data to samples if it doesn't add itself to channel meta

### LINKED ISSUE #1062: LapDelta channel added to samples when not added to channel_metas
The LaptimeDeltaFilter (and probably other filters) still add their channels to samples from RCP even if the channels they rely upon don't exist. This breaks code that relies upon the channel metas to be the source of truth of what channels will show up in a sample.

For example, LaptimeDeltaFilter won't add its meta to the channel meta:

```
    def get_channel_meta(self, channel_meta):
        metas = {}
        if channel_meta.get(self.LAPTIME_KEY):
            metas[LaptimeDeltaFilter.LAP_DELTA_KEY] = self.lap_delta_meta
        return metas
```

But it will add its value(s) in `filter`:

```
    def filter(self, channel_data):
        laptime = channel_data.get(LaptimeDeltaFilter.LAPTIME_KEY)
        predtime = channel_data.get(LaptimeDeltaFilter.PREDTIME_KEY)
        best_laptime = channel_data.get(LaptimeDeltaFilter.BEST_LAPTIME_KEY)
        
        #we choose predicted laptime if it is present in the data stream
        reference_laptime = laptime if predtime == None else predtime
        delta_time = 0.0
        if reference_laptime and best_laptime:
            delta_time = (reference_laptime - best_laptime) * 60.0
        channel_data[LaptimeDeltaFilter.LAP_DELTA_KEY] = delta_time
```

ryandoherty:
Fixes #1062


autosportlabs/racecapture/databus/filter/laptimedeltafilter.py
```diff
@@ -1,3 +1,23 @@
+#
```
brentpicasso: Please use the same license language in the root of the project :+1:

ryandoherty: Huh this was copy-pasted from CONTRIBUTING.md? Is there another template?

brentpicasso: I was referring to the LICENSE file. Maybe we should normalize on that, including updating the CONTRIBUTING.md text.

brentpicasso:
:shipit: after resolving license text.

brentpicasso:
:shipit:  I didn't read license blurb carefully enough :)

PR #1081, opened 2016-07-20
Streaming api support

ryandoherty:
Adds streaming api support over wifi/usb for sample data.

### DIFF
--- a/autosportlabs/racecapture/databus/databus.py
+++ b/autosportlabs/racecapture/databus/databus.py
@@ -7,6 +7,7 @@ from autosportlabs.racecapture.data.sampledata import Sample, SampleMetaExceptio
 from autosportlabs.racecapture.databus.filter.bestlapfilter import BestLapFilter
 from autosportlabs.racecapture.databus.filter.laptimedeltafilter import LaptimeDeltaFilter
 from autosportlabs.util.threadutil import safe_thread_exit
+from autosportlabs.racecapture.config.rcpconfig import Capabilities
 from utils import is_mobile_platform
 
 DEFAULT_DATABUS_UPDATE_INTERVAL = 0.02  # 50Hz UI update rate
@@ -200,6 +201,7 @@ class DataBusPump(object):
 
     def __init__(self, **kwargs):
         super(DataBusPump, self).__init__(**kwargs)
+        self.rc_capabilities = None
 
     def start(self, data_bus, rc_api, streaming_supported):
         if self._running.is_set():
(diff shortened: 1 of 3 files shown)


autosportlabs/racecapture/databus/databus.py
```diff
@@ -214,19 +216,45 @@ def start(self, data_bus, rc_api, streaming_supported):
...
+        Logger.debug("DataBusPump: Checking if new streaming api is supported")
+
+        def handle_capabilities(capabilities_dict):
+            self.rc_capabilities = Capabilities()
+            self.rc_capabilities.from_json_dict(capabilities_dict['capabilities'])
+
+            if self.rc_capabilities.has_streaming:
+                # Send streaming command
+                Logger.debug("DataBusPump: connected device supports streaming")
+                self._rc_api.start_telemetry(50)
```
brentpicasso: let's extract this number. suggest setting the number at the top, and then derive the 0.02 from this number (1/50) so the same number drives both rates. :+1:

autosportlabs/racecapture/databus/databus.py
```diff
@@ -214,19 +216,45 @@ def start(self, data_bus, rc_api, streaming_supported):
...
+            else:
+                Logger.debug("DataBusPump: connected device does not support streaming api, starting polling")
+                t = Thread(target=self.sample_worker)
+                t.start()
+                self._sample_thread = t
+
+        def handle_capabilities_fail():
+            Logger.error("DataBusPump: Failed to get capabilities, can't determine if device supports streaming API")
```
brentpicasso: Seeing this made me think, maybe we can log an error through the crash handler, without blowing an exception up the stack?

brentpicasso: Not critical, just a thought.

brentpicasso:
Found an issue where if I unplug / replug RCT or RCP the streaming is not re-activated.

Otherwise, this looks pretty solid so far!

autosportlabs/racecapture/databus/databus.py
ryandoherty: 👍

autosportlabs/racecapture/databus/databus.py
ryandoherty: Hmm yeah I was thinking about this too, it is technically an error. I will look into Sentry and see if we can log this to Sentry

ryandoherty:
Made fixes per comments, in order to fix the reconnect issue I need some code from the autologging PR, will wait on that to be merged then fix this one.

ryandoherty:
Ok @brentpicasso , reconnect logic added and tested over USB. having wifi issues with RCT, haven't tested yet.

brentpicasso:
Tested, works, :shipit:

PR #1086, opened 2016-07-25
Wifi mobile

### LINKED ISSUE #1082: UDP discovery not working on android
Not getting UDP ping from RC, but desktop seems to work.

brentpicasso:
Gets wifi working on mobile platform. 

resolves #1082

### DIFF
--- a/autosportlabs/comms/socket/socketconnection.py
+++ b/autosportlabs/comms/socket/socketconnection.py
@@ -44,7 +44,7 @@ class SocketConnection(object):
         sock.settimeout(SCAN_TIMEOUT)
 
         # Bind the socket to the port
-        server_address = ('0.0.0.0', PORT)
+        server_address = ('', PORT)
         sock.bind(server_address)
 
         try:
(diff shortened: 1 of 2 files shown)


autosportlabs/comms/socket/socketconnection.py
```diff
@@ -44,7 +44,7 @@ def get_available_devices(self):
         sock.settimeout(SCAN_TIMEOUT)
 
         # Bind the socket to the port
-        server_address = ('0.0.0.0', PORT)
+        server_address = ('', PORT)
```
ryandoherty: This is weird, because '0.0.0.0' means 'bind to all interfaces', so I'm wondering why it needs to be changed. But the Python docs to say that a blank host string means bind to all interfaces. :shrug:?

ryandoherty:
lgtm, wondering about the bind address thing but not going to sweat it.

autosportlabs/comms/socket/socketconnection.py
brentpicasso: Yeah, I was wondering about this as well.

Issue #1102, opened 2016-08-01
For controlling telemetry, use "setTelemetry" method


stieg:
I realize now that the `setTelemetryStart` and `setTelemetryStop` methods are a bit redundant.  Rather, by changing a default value I can simply consolidate them into one API method: `setTelemetry`.  This patch will be out for review shortly and I would like the final version of 2.10.0 to only have this method and hence be able to remove the `setTelemetryStart` and `setTelemetryStop` methods.  The way you control it is to set a rate of <= 0 to stop the telemetry stream, else a valid telemetry sample rate value otherwise.  Once this is fixed I can resolve https://github.com/autosportlabs/RaceCapture-Pro_firmware/issues/785 and then we will be solid for release.

PR #1103, opened 2016-08-02
Migrate to new setTelemetry API method

stieg:
Because the old setTelemetry{Start,Stop} was redundant. Issue #1102


ryandoherty:
Assuming this has been tested with 2.10.0, then :shipit: !

stieg:
This is now tested and validated on 2.10.0.  Will be fully operational when https://github.com/autosportlabs/RaceCapture-Pro_firmware/issues/787 is merged.

Issue #1122, opened 2016-08-15
App fails to flush serial device before writing/reading "getVer" probe


stieg:
It seems that our SocketComm has some messages queued up that prevent it from reading the correct JSON reply.  See the following snippet of the log below:

```
[DEBUG             ] [SocketComm  ] writing message {"getVer":null}
[DEBUG             ] [SocketConnection] returning data {"s":{"t":346200,"d":[346202,0,0.22,0.13,0.85,-0.79,-3.95,2.73,0.0,0.0,0.0,0.0,0.0,0,0,0.0,0,0.0,-1,0.0,0.0,0.0,0,16777211]}}
```

Notice how we are reading sample data instead of the reply?  We should try to read multiple messages to see if its the correct response or should flush the channel before we read it to ensure there are no messages in there.  This will help with the reconnect issue we are seeing.  Full log below:

```
[DEBUG             ] [SocketConnection] returning data {"s":{"t":346080,"d":[346082,0,0.22,0.13,0.85,-0.79,-3.47,2.49,507]}}
[DEBUG             ] [SocketConnection] returning data {"s":{"t":346100,"d":[346102,0,0.0,0.0,0.0,0.0,0.0,0,0,0.0,0,0.0,-1,0.0,0.0,0,14679555]}}
[WARNING           ] [RCPAPI      ] Read message timeout waiting for capabilities
[WARNING           ] [RCPAPI      ] POKE
[WARNING           ] [RCPAPI      ] Level 2 retry for (3) capabilities
[ERROR             ] [RCPAPI      ] Command sequence exception: Timeout waiting for capabilities
[ERROR             ] [Traceback (most recent call last)]
  File "/home/stieg/Devel/asl/app/autosportlabs/racecapture/api/rcpapi.py", line 331, in cmd_sequence_worker
    raise Exception('Timeout waiting for ' + name)
Exception: Timeout waiting for capabilities

[ERROR             ] [Main        ] Timeout waiting for capabilities
[INFO              ] [TelemetryManager] on_data_connected: False
[DEBUG             ] [TelemetryManager] stop()
[INFO              ] [SessionRecorder] stopping session
[DEBUG             ] [RCPAPI      ] attempting to recover connection
[DEBUG             ] [RCPAPI      ] Execute Sequence complete
[DEBUG             ] [RCPAPI      ] Starting auto-detect
[DEBUG             ] [SocketComm  ] close()
[DEBUG             ] [SocketComm  ] closing connection process
[...]

PR #1123, opened 2016-08-15
Double the time we wait for the version message

stieg:
Waiting 1 second doesn't seem to be enough when it comes to reading
the version information that gets returned from the RCT device.
This was preventing us from reconnecting reliably.  Bumping this to
two seconds made it much more reliable.

Issue #1122


ryandoherty:
Looks good to me, guessing this is a wifi problem or with all connections?

stieg:
Don't know... only working on WiFi at the moment

PR #1132, opened 2016-08-17
add listeners after values are set to prevent incorrect overwriting o…

### LINKED ISSUE #1108: OBDII config not updating sample rate when reading config
When reading the config, the OBDII view does not correctly update the sample rate.

STR

add OBDII channel
set sample rate to 10
write config
read config
observe sample rate is disabled

when running debug mode, firmware does send back proper sample rate value:
[DEBUG  ] [RCPAPI      ] Tx: {"getObd2Cfg":null}
[DEBUG  ] [RCPAPI      ] Rx: {"obd2Cfg":{"en":1,"pids":[{"nm":"TPS","ut":"%","min":0,"max":100,"prec":0,"sr":10,"pid":17}]}}

brentpicasso:
…f values

resolves #1108

### DIFF
--- a/autosportlabs/racecapture/views/configuration/rcp/obd2channelsview.kv
+++ b/autosportlabs/racecapture/views/configuration/rcp/obd2channelsview.kv
@@ -12,7 +12,6 @@
     SampleRateSpinner:
     	size_hint_x: 0.4
     	id: sr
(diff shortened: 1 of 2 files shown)


ryandoherty:
Looks good to me.

PR #1133, opened 2016-08-18
Toast when session starts/stops

### LINKED ISSUE #1093: SessionRecorder: Toast when session recording starts and stops.
Show a Toast when a session starts recording, and when a session stops recording, so the user knows it happened.

brentpicasso:
resolves #1093

### DIFF
--- a/main.py
+++ b/main.py
@@ -51,6 +51,7 @@ if __name__ == '__main__':
     from autosportlabs.help.helpmanager import HelpInfo
     from autosportlabs.racecapture.datastore import DataStore
     from autosportlabs.racecapture.data.sessionrecorder import SessionRecorder
+    from autosportlabs.uix.toast.kivytoast import toast
     from toolbarview import ToolbarView
     if not is_mobile_platform():
         kivy.config.Config.set ('input', 'mouse', 'mouse,multitouch_on_demand')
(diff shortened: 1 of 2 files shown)


main.py
```diff
@@ -529,6 +531,8 @@ def _restart_comms(self):
         self._rc_api.shutdown_api()
         self.init_rc_comms()
 
+    def _on_session_recording(self, instance, is_recording):
+        toast('Analysis session started' if is_recording else 'Analysis session stopped', length_long=True)
```
ryandoherty: The messaging here seems unclear to me. What is an 'analysis session'? I'd go with something like 'recording' or 'logging' session instead, seems clearer to me.

ryandoherty:
Looks good other than wordsmithing!

main.py
brentpicasso: Logging might be a good one. I don't want to confuse that with RCP SD card logging, though.

I'll try "Recording Session' or similar.

ryandoherty:
Soooooooooooooo I hate to be that guy, but I think wording that explicitly says 'recording' would make it very self-explanatory. I don't think users will understand that 'Session started' and 'Session ended' means the app is recording. We could make it 'Session recording started' and 'Session recording ended', which to me seems much clearer. I do know that we have to make sure it's clear it isn't SD card recording though.

ryandoherty:
:shipit:

PR #1135, opened 2016-08-18
default to all geo regions

brentpicasso:
Have the track selection widget default to all geo regions instead of west coast. b/c polite.


ryandoherty:
:shipit:

PR #1139, opened 2016-08-18
removed 20% font scaling for BetterTextInput

### LINKED ISSUE #1113: WiFi config - Access Point SSID / Password fields need scaling
The height percentage needs to be scaled in a similar manner as the client mode fields.

The effect shows the most when viewed on a phone

brentpicasso:
resolves #1113


ryandoherty:
LGTM!

PR #1142, opened 2016-08-19
Validate ip address when opening connection

ryandoherty:
STR:

* Open app, connect to RCT or RCP with wifi via USB
* Turn on WiFi module, enable AP mode
* Close app
* Connect to AP with your computer
* Run app with `--conn_type=wifi` cli arg
* Watch as the app tries to connect to something in `/dev/`, and it will hang for a while, then (usually) never connect to RC
* Eventually it will throw an exception after ~10s

This happens because the app uses the last connection when trying to connect again to RC, but `/dev/foo` is not a valid IP. For some reason the Python socket lib takes ages to throw an error, so autodetection appears to fail. To fix it, validate IP addresses before connecting and throw an error if it isn't valid.

### DIFF
--- a/autosportlabs/comms/socket/socketcomm.py
+++ b/autosportlabs/comms/socket/socketcomm.py
@@ -34,7 +34,7 @@ class SocketComm(object):
     Responsible for communicating with a socket connection to a RC device
      Starts up threads for reading/writing data and pushes/reads data to queues
     """
-    CONNECT_TIMEOUT = 1.0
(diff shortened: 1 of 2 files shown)


ryandoherty:
Note: this only happens when the phone/tablet/computer is connected to RC's AP, not when RC is in client mode.

brentpicasso:
tight, clean solution. :shipit:

PR #1152, opened 2016-08-19
Changes all of our commands to end with \r\n instead of \r

stieg:
Because we use both serial line and non-serial lines to read/parse commands we should probably be consistent and use a consistent serial message EOF delimiter \r\n.  This makes things consistent with telnet which does this, and thus gives us a consistent way to access our device.  tested on RCT over WiFi and USB.  Have yet to test BT.


stieg:
BAD PR.  Closing

PR #1153, opened 2016-08-19
Delimit all commands with \r\n instead of \r

stieg:
This is becasue we treat all commands as serial line commands.  Thus
we should use proper serial line delimeters to delimit the end of a
command.

As an added bonus this helps us work around a shortcoming in the ESP8266
chip (IPD doesn't terminate lines with \r\n).  This allows our serial
library to always read and handle the line appropriately.

### DIFF
--- a/autosportlabs/racecapture/api/rcpapi.py
+++ b/autosportlabs/racecapture/api/rcpapi.py
@@ -68,6 +68,7 @@ class RcpApi:
     _auto_detect_busy = Event()
 
     COMMAND_SEQUENCE_TIMEOUT = 1.0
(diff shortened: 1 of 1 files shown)


ryandoherty:
LGTM! Will this work with firmware < 2.10.0?

brentpicasso:
Please re-open against 1.7.1 after regression testing against MK1 and MK2 (older firmware versions)

ryandoherty:
Just tested this with 2.8.4/7 on Mk1 and 2.9.2 and 2.10.0 on Mk2, works fine. Safe to merge into 1.7.7

PR #1154, opened 2016-08-20
Update min firmware version number

ryandoherty:
So many changes to the API, need to require 2.10.0


brentpicasso:
What does this do for MK1 support?

brentpicasso:
I checked MK1 2.8.x and top of 1.7.0 seems to read / write correctly save for #1158 and #1156 

Recommend not merging this PR.

ryandoherty:
WFM!

PR #1161, opened 2016-08-21
Update flags for links capabilities, fixes #1158

### LINKED ISSUE #1158: flag dependent config screens don't correctly refresh after config is loaded
Specifically, wireless config screen. 

If you navigate to the screen, it loads a default view with all options enabled. Once the config is read,  sections are not refreshed to reflect actual capabilities for MK2 and MK1.  RCT looks correct.

If you navigate to the screen after the config is loaded, incorrect options are shown.
MK2: 
No cellular or BT configuration shown
MK1:
Screen is simply blank

ryandoherty:
Fixes #1158

### DIFF
--- a/autosportlabs/racecapture/config/rcpconfig.py
+++ b/autosportlabs/racecapture/config/rcpconfig.py
@@ -1100,8 +1100,8 @@ class LinksCapabilities(object):
         self.usb = True
 
     def from_flags(self, flags):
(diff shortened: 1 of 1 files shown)


brentpicasso:
This fixes part of #1158. 

MK2 2.10.0 behaves correctly;  
MK2 with 2.9.2 and MK1 2.8.7 gives a wireless page that is blank.

PR #1164, opened 2016-08-21
ensure we have version config info so we know what to query for when …

brentpicasso:
…we read the device's configuration

Resolves remaining issues for #1158

### DIFF
--- a/autosportlabs/racecapture/config/rcpconfig.py
+++ b/autosportlabs/racecapture/config/rcpconfig.py
@@ -1101,10 +1101,10 @@ class StorageCapabilities(object):
 class LinksCapabilities(object):
 
     def __init__(self):
-        self.bluetooth = True
-        self.cellular = True
-        self.wifi = True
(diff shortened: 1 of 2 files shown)


autosportlabs/racecapture/config/rcpconfig.py
```diff
@@ -1101,21 +1101,22 @@ def to_json_dict(self):
...
 
     def __init__(self):
-        self.bluetooth = True
-        self.cellular = True
-        self.wifi = True
-        self.usb = True
+        self.bluetooth = False
+        self.cellular = False
+        self.wifi = False
+        self.usb = False
```
ryandoherty: So this will mean that if you don't have anything plugged in the Wireless config view will be blank?

ryandoherty: Oops nvm.

ryandoherty:
LGTM :shipit:

PR #1166, opened 2016-08-22
Make session recorder start() and stop() idempotent

### LINKED ISSUE #1145: "Session recording stopped" toast appears during firmware update
Happens even if I haven't gone to the Dashboard view, which is what starts a recording session.

ryandoherty:
Fixes #1145


brentpicasso:
:shipit:

PR #1175, opened 2016-08-25
Fix IMU channel spinner resetting to default on every other config read

### LINKED ISSUE #1173: IMU view selects wrong channel label on every other read config
Wat

https://youtu.be/rRMK1X9ClBA

ryandoherty:
Fixes #1173 

Somehow when setImuType is called and setValueMap is called more than once, it resets the channel config (probably because it's changing variables that are references).


brentpicasso:
Looks good, merging!

PR #1176, opened 2016-08-25
Reread accel config after calibration

ryandoherty:
After calibration, we need to re-read the configuration so the app has the latest values. Added some state to prevent the app thinking the configuration had changed when we had actually just updated from RC's configuration.


brentpicasso:
looks good! merging.

PR #1179, opened 2016-08-26
Add state to DataBusPump to prevent multiple setTelemetry calls to RC

### LINKED ISSUE #1104: App Sends Duplicate `setTelemetry` Calls to Start a Telemetry Stream
On top of tree 1.7.0 if the telemetry stream is running and I close the app then, the stream is not stopped.  While this should not cause an issue on the firmware, it currently does (https://github.com/autosportlabs/RaceCapture-Pro_firmware/issues/786).  The app should be a good citizen and should stop the stream upon exit.

ryandoherty:
Fixes #1104 . Because the DataBusPump can be started in multiple ways and has to wait for RC responses to figure out how to start fetching data, it's possible it may send multiple `setTelemetry` calls to RC. I added some state so if it is in the middle of querying RC and is told to start repeatedly, it will do the appropriate thing and not query repeatedly.

### DIFF
--- a/autosportlabs/racecapture/databus/databus.py
+++ b/autosportlabs/racecapture/databus/databus.py
@@ -204,9 +204,16 @@ class DataBusPump(object):
         self.rc_capabilities = None
         self._should_run = False
         self._running = False
+        self._starting = False
         self._streaming_supported = False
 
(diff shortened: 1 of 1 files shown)


brentpicasso:
looks good, merging.

Issue #1112, opened 2016-08-11
Session browser drop down sometimes is added twice


brentpicasso:
Under certain conditions, the session browser in analysis view can appear twice in the analysis view. 

Still reproducing, but seems to coincide with session recording, or if a session was previously recorded. 

Another hint is from the main screen, a tap on the analysis button resulted in no activity.  Then, a 2nd tap resulted in navigating to the analysis page, and two session browsers were visible.

PR #1200, opened 2016-09-07
prevent multiple layouts from loading in analysis view, including mul…

### LINKED ISSUE #1187: Session drop down created multiple times / multiple event handlers
Need to make the initialization of the analysis page idempotent - so multiple widgets and event handlers are not created if it is initialized more than once.

brentpicasso:
…tiple flyin handles; prevent error when no trackmap geopoints are available; prevent premature dismissing of session flyin panel

resolves #1187 , #1112 

sorry for the whitespace changes, my editor auto-cleaned the files.


autosportlabs/racecapture/geo/geopoint.py
```diff
@@ -2,17 +2,31 @@
 RADIUS_EARTH_KM = 6371;
 
 class GeoPoint:
+    '''
```
ryandoherty: Just an FYI that the PEP8 standard is double quotes, not single. https://www.python.org/dev/peps/pep-0257/

autosportlabs/racecapture/views/analysis/flyinpanel.py
```diff
@@ -64,41 +65,41 @@ class FlyinPanelException(Exception):
...
+
     def __init__(self, **kwargs):
         Builder.load_string(FLYIN_PANEL_LAYOUT)
         super(FlyinPanel, self).__init__(**kwargs)
         self.hide_decay = Clock.create_trigger(lambda dt: self.hide(), self.SESSION_HIDE_DELAY)
         Window.bind(mouse_pos=self.on_mouse_pos)
         Window.bind(on_motion=self.on_motion)
         Clock.schedule_once(lambda dt: self.show())
-    
+        self.shown_at = None
```
ryandoherty: Seems like this is a private var? If so probably best to prefix with `_`.

brentpicasso: will do

autosportlabs/racecapture/geo/geopoint.py
brentpicasso: fixed here and elsewhere.

Issue #1149, opened 2016-08-19
Exception when attempting to connect to ip address via serial comms


stieg:
After using the app in wifi mode on my laptop I see these exceptions.  Looks like a bug in how we are trying to open our serial based on the previous device that worked.

```
[INFO              ] [RacecaptureApp] initializing rc comms with, conn type: None
[INFO              ] [RCPAPI      ] msg_rx_worker starting
[INFO              ] [RCPAPI      ] cmd_sequence_worker starting
[DEBUG             ] [RCPAPI      ] Port not open...
[INFO              ] [RCPAPI      ] auto_detect_worker starting
[DEBUG             ] [ImagePygame ] Load <resource/images/app_icon_512x512.png>
[DEBUG             ] [RCPAPI      ] Starting auto-detect
[DEBUG             ] [SerialConnection] getting available devices
[INFO              ] [RCPAPI      ] trying last known device first: 192.168.1.252
[DEBUG             ] [RCPAPI      ] Searching for device
[DEBUG             ] [RCPAPI      ] Trying 192.168.1.252
[DEBUG             ] [Comms       ] Opening connection 192.168.1.252
[INFO              ] [RaceCaptureApp] Current Tracks Loaded
[DEBUG             ] [RCPAPI      ] Tx: {"getVer":null}
[DEBUG             ] [Comms       ] connection process starting
[DEBUG             ] [Comms       ] Exception setting up connection process: <class 'serial.serialutil.SerialException'>[Errno 2] could not open port 192.168.1.252: [Errno 2] No such file or directory: '192.168.1.252'
[DEBUG             ] [Traceback (most recent call last)]
  File "/home/stieg/Devel/asl/app/autosportlabs/comms/comms.py", line 48, in connection_message_process
    connection.open(device)
  File "/home/stieg/Devel/asl/app/autosportlabs/comms/serial/serialconnection.py", line 30, in open
    self.ser = serial.Serial(device, timeout=self.timeout, write_timeout=self.writeTimeout)
  File "/home/stieg/Devel/asl/app/fedora_env/lib/python2.7/site-packages/serial/serialutil.py", line 171, in __init__
    self.open()
  File "/home/stieg/Devel/asl/app/fedora_env/lib/python2.7/site-packages/serial/serialposix.py", line 294, in open
    raise SerialException(msg.errno, "could not open port %s: %s" % (self._port, msg))
[...]

PR #1204, opened 2016-09-08
make serial comms error quieter in log

brentpicasso:
Addresses #1149.   exceptions, especially serial exceptions are common, especially when the available list of port changes, no need  to print stack trace.


ryandoherty:
Hmm, could we change this to trace or something similar? Since we're catching all Exceptions here we might end up catching other codes' errors, having a stack trace the first time makes fixing those errors much easier.

brentpicasso:
Sure, we can make the stack trace as trace. Will do that.

brentpicasso:
Ok, changed to trace.

PR #1207, opened 2016-09-09
Remove delay when clicking/tapping 'Send telemetry' button in app preferences

### LINKED ISSUE #1184: Responsiveness of BT <-> wifi switch
The touch response seems poor for this button.  Investigate why and resolve.

Possible reasons: UI thread is blocked when the switch is activated.

ryandoherty:
Seems to work better now, UI still shows a blue background color for a second, which we have no control over afaik. 

Fixes #1184

### DIFF
--- a/main.py
+++ b/main.py
@@ -542,14 +542,20 @@ class RaceCaptureApp(App):
             if value == "1":  # Boolean settings values are 1/0, not True/False
                 if self.rc_config.connectivityConfig.cellConfig.cellEnabled:
                     alertPopup('Telemetry error', "Turn off RaceCapture's telemetry module for app to stream telemetry.")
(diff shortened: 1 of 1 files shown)


brentpicasso:
#1184 actually is a different problem, observed when switching from BT to WiFi (app seems to hang for a while).  I've also observed this one as a separate issue (button responsiveness) but I don't think we're tracking an issue.

ryandoherty:
Oh wow, I confused the two, oops. This change is still nice though! :P

brentpicasso:
Yeah totally important!  I'll review in the meantime.  Should get it's own issue too. :)

brentpicasso:
:shipit:

PR #1208, opened 2016-09-09
Dynamically adjust telemetry speed

### LINKED ISSUE #1190: Adjust telemetry rate based on app mode
When in dashboard mode (or logging a session) - set telemetry rate to 50Hz
When in other modes (and not logging a session), set telemetry rate to 1Hz

This lets data periodically update for various monitoring purposes, and ups the rate to full logging rate when in dashboard / session logging mode.

brentpicasso:
resolves #1190

### DIFF
--- a/autosportlabs/racecapture/api/rcpapi.py
+++ b/autosportlabs/racecapture/api/rcpapi.py
@@ -209,10 +209,7 @@ class RcpApi:
                     msg = unicode(msg, errors='ignore')
                     msgJson = json.loads(msg, strict=False)
 
-                    if 's' in msgJson:
-                        Logger.trace('RCPAPI: Rx: ' + str(msg))
-                    else:
-                        Logger.debug('RCPAPI: Rx: ' + str(msg))
+                    Logger.debug('RCPAPI: Rx: ' + str(msg))
                     Clock.schedule_once(lambda dt: self.on_rx(True))
                     error_count = 0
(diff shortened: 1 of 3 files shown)


autosportlabs/racecapture/api/rcpapi.py
```diff
@@ -209,10 +209,7 @@ def msg_rx_worker(self):
                     msg = unicode(msg, errors='ignore')
                     msgJson = json.loads(msg, strict=False)
 
-                    if 's' in msgJson:
-                        Logger.trace('RCPAPI: Rx: ' + str(msg))
-                    else:
-                        Logger.debug('RCPAPI: Rx: ' + str(msg))
+                    Logger.debug('RCPAPI: Rx: ' + str(msg))
```
ryandoherty: It would be nice if there was some way to avoid seeing all telemetry data in debug mode because it's a firehose and if you are debugging other transmitted information, it's pretty much impossible to see anything.

ryandoherty:
Overall looks good! If there is a way to write a unit test(s) for this behavior I would do it.

autosportlabs/racecapture/api/rcpapi.py
brentpicasso: agreed. re-enabled it.

brentpicasso:
agree on the unit test. Will have to handle later. :-/

PR #1219, opened 2016-09-12
allow touch events to reach linechart

### LINKED ISSUE #1217: Touch events are not propogating on the right side of the line chart
The ChannelsValueView seems to be gobbling the touch events. you can only swipe into that region when you start from the left side. Need to be able to propagate touch events to the lower widgets.

brentpicasso:
resolves #1217 

A simple fix, almost just pushed it, thought might be helpful to share though.


ryandoherty:
:shipit: ! 👍

PR #1220, opened 2016-09-12
Incrementally load session and selected laps, to improve UI responsiveness

### LINKED ISSUE #1213: analysis view loads slowly under certain conditions
If a lap of a session is pre-selected and the lap data is very large, the loading of the analysis screen will be stalled, and app will appear to hang. 

The screen should load first, then perform any data querying after the screens are loaded.

brentpicasso:
Resolves #1213 

before/after with sonoma data set and 5 laps selected

Nvidia Shield tablet:
Before: 6 second hang before showing screen after touching analysis button on home page
After: < 1 second after touching analysis button

Samsung S5:
Before: 12 second hang before initial screen shows
After: < 1 second after touching analysis button

### DIFF
--- a/autosportlabs/racecapture/views/analysis/analysisview.py
+++ b/autosportlabs/racecapture/views/analysis/analysisview.py
@@ -120,8 +120,8 @@ class AnalysisView(Screen):
         self.stream_connecting = False
         Window.bind(mouse_pos=self.on_mouse_pos)
         Window.bind(on_motion=self.on_motion)
-        self.init_view()
-
+        self._layout_complete = False
+        
     def on_motion(self, instance, event, motion_event):
         flyin = self.ids.laps_flyin
         if self.collide_point(motion_event.x, motion_event.y):
@@ -261,6 +261,12 @@ class AnalysisView(Screen):
         self.ids.sessions_view.init_view()
         Clock.schedule_once(lambda dt: HelpInfo.help_popup('beta_analysis_welcome', self, arrow_pos='right_mid'), 0.5)
 
+    def do_layout(self, *largs):
+        super(AnalysisView, self).do_layout(largs)
+        if not self._layout_complete:
+            Clock.schedule_once(lambda dt: self.init_view(), 0.5)
+        self._layout_complete = True
+        
     def _init_datastore(self):
         dstore_path = self._settings.userPrefs.datastore_location
         if os.path.isfile(dstore_path):
--- a/autosportlabs/racecapture/views/analysis/sessionlistview.py
+++ b/autosportlabs/racecapture/views/analysis/sessionlistview.py
@@ -154,11 +154,13 @@ class SessionListView(AnchorLayout):
             selection_settings = json.loads(selection_settings_json)
             delete_sessions = []
 
+            session_selections = []
+
             # Load sessions first, then select the session
             for session_id_str, session_info in selection_settings["sessions"].iteritems():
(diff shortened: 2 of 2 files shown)


autosportlabs/racecapture/views/analysis/analysisview.py
```diff
@@ -261,6 +261,12 @@ def init_view(self):
         self.ids.sessions_view.init_view()
         Clock.schedule_once(lambda dt: HelpInfo.help_popup('beta_analysis_welcome', self, arrow_pos='right_mid'), 0.5)
 
+    def do_layout(self, *largs):
```
ryandoherty: Where/how is this called? Is it a Kivy thing?

brentpicasso: It overrides a kivy function.  It's the latest point I could identify where
the layout is completely rendered.

ryandoherty: Oh hey lgmtfy, yes it is a Kivy thing.

autosportlabs/racecapture/views/analysis/sessionlistview.py
```diff
@@ -170,18 +171,29 @@ def init_view(self):
...
+                    lap_selections.append((session_id, lap))
+            Clock.schedule_once(lambda dt: self._load_next_selected_session(0, session_selections, lap_selections), 0.1)
         else:
             Logger.error("SessionListView: init_view failed, missing settings or datastore object")
             raise Exception("SessionListView: init_view failed, missing settings or datastore object")
 
+    def _load_next_selected_session(self, index, session_selections, lap_selections):
+        if index < len(session_selections):
+            self.append_session(session_selections[index])
+            Clock.schedule_once(lambda dt: self._load_next_selected_session(index + 1, session_selections, lap_selections), 0.1)
```
ryandoherty: I tested not giving schedule_once any delay, seems to be faster and the app still works. That would speed up the load time by 0.1s for each session and lap, so probably on average 0.5-1s. Seems like it's safe to do, thoughts?

brentpicasso: Try it both ways on a slower device (phone / tablet). With the 0.1 the
provides more visual update activity,  which imo is a better experience.  I
think the 0 delay might cause the ui to visually hang.  Maybe 0.05?

ryandoherty:
Have we profiled this to figure out what pieces are causing the slowdown and how it could be optimized?

brentpicasso:
Did not re-profile since the last time. I am currently assuming it is reasonably optimized based on the last round of optimizations we did by moving the sqlite querying into the UI thread.  

I think it makes sense to spread the work out because the user can still select many laps, and you could still have a long load time with no UI responsiveness, even if it is further optimized than what we did last time.

autosportlabs/racecapture/views/analysis/sessionlistview.py
ryandoherty: Tried it with no delay, 0.05 and 0.1. 0.1 did work the best it seemed. I think 0.05 is too fast and the app gets bogged down.

ryandoherty:
Overall code looks fine, we should ship this. Medium term we should profile this deeper to figure out exactly what pieces of code and/or database activity is slow.

brentpicasso:
I agree. I think some DB optimizations can be had here.  Thanks for the review!

PR #1221, opened 2016-09-13
filter for only valid laps. (remove -1 lap)

### LINKED ISSUE #1211: remove -1 lap in session list
Should not show the -1 lap in the session list.

brentpicasso:
resolves #1211

I wanted to do this solely through modifying the query, but the two part query seemed like it would be cleaner to filter it at the end .


ryandoherty:
:shipit: , don't forget to update the changelog :)

brentpicasso:
thanks. will do after merging

PR #1237, opened 2016-09-15
P/brent/map laptime

### LINKED ISSUE #1066: Show laptime in analysis view lap list
When I'm reviewing 2 or more laps, it helps to have the laptimes visible once the Sessions dropdown slides back up. Otherwise it's hard to tell which lap was faster when comparing very similar laps:

http://i.imgur.com/nm3lyBr.png

brentpicasso:
resolves #1066 and includes other refactoring and improvements.


autosportlabs/racecapture/views/analysis/analysisdata.py
```diff
@@ -49,6 +49,20 @@ def __init__(self, **kwargs):
...
+        """
+        Provides a cached representation of the list of sessions and laps.
+        @return dict of Session objects
+        """
+        session_info = self._session_info_cache
+
+        # Populate our cache if necessary
+        if len(session_info.keys()) == 0:
+            self._refresh_session_data()
+        return session_info
```
ryandoherty: I'm curious why `self._session_info_cache` is assigned to `session_info` and then returned. Seems like this could be a little simpler and just be:

```
if len(self._session_info_cache.keys()) == 0:
    self._refresh_session_data()

return self._session_info_cache
```

autosportlabs/racecapture/views/analysis/channelvaluesview.kv
```diff
@@ -27,14 +27,15 @@
...
+        halign: 'right'
     FieldLabel:
-        size_hint_x: 0.1
+        size_hint_x: 0.4
         id: lap
-    FieldLabel:
-        size_hint_x: 0.40
-        id: channel
+        halign: 'right'
+        text: ' {} :: {}  '.format(root.session, root.lap)
```
ryandoherty: I feel like this probably should belong in the .py file, mostly because that's where people will expect it to be if it needs to be changed. Could get confusing.

brentpicasso: I picked this because I've seen it elsewhere from kivy examples where the automatic bindings of  StringProperties occur - looked like a best practice from Kivy. Plus the formatting is in the kivy layout; IMO it works b/c the the formatting is closer to the presentation / layout.

autosportlabs/racecapture/views/analysis/analysisdata.py
brentpicasso: As a general practice, if I reference a member variable more than once in a function I usually assign it to a local variable first before doing things with it. local variables are significantly faster (33%) and encapsulates access to that member variable.

ryandoherty: Ok 👍

PR #1239, opened 2016-09-15
automatically show add session dialog if no sessions are selected

### LINKED ISSUE #1238: if we load analysis views with no sessions in session list, show add session dialog immediately
We should not have users be faced with a blank analysis window with nothing showing. If they enter the screen with no sessions loaded, they should be shown the add session dialog

brentpicasso:
to be helpful and discoverable. resolves #1238


ryandoherty:
:shipit: !

PR #1242, opened 2016-09-15
add preference to auto-record session

### LINKED ISSUE #1240: Provide preferences switch to enable in-app session recording
Switch should default to 'on'

brentpicasso:
Resolves #1240

### DIFF
--- a/autosportlabs/racecapture/settings/prefs.py
+++ b/autosportlabs/racecapture/settings/prefs.py
@@ -173,8 +173,9 @@ class UserPrefs(EventDispatcher):
         self.config.setdefault('preferences', 'config_file_dir', default_user_files_dir)
         self.config.setdefault('preferences', 'firmware_dir', default_user_files_dir)
         self.config.setdefault('preferences', 'import_datalog_dir', default_user_files_dir)
-        self.config.setdefault('preferences', 'first_time_setup', True)
-        self.config.setdefault('preferences', 'send_telemetry', False)
+        self.config.setdefault('preferences', 'first_time_setup', '1')
+        self.config.setdefault('preferences', 'send_telemetry', '0')
+        self.config.setdefault('preferences', 'record_session', '1')
         self.config.setdefault('preferences', 'last_dash_screen', 'gaugeView')
         self.config.setdefault('preferences', 'global_help', True)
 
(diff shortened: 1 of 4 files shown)


autosportlabs/racecapture/settings/prefs.py
```diff
@@ -173,8 +173,9 @@ def set_config_defaults(self):
         self.config.setdefault('preferences', 'config_file_dir', default_user_files_dir)
         self.config.setdefault('preferences', 'firmware_dir', default_user_files_dir)
         self.config.setdefault('preferences', 'import_datalog_dir', default_user_files_dir)
-        self.config.setdefault('preferences', 'first_time_setup', True)
-        self.config.setdefault('preferences', 'send_telemetry', False)
```
ryandoherty: Why the change from boolean to strings? Also, I think this should have gone into a separate PR, seems unrelated.

resource/settings/settings.json
```diff
@@ -10,6 +10,14 @@
     },
     {
         "type": "bool",
+        "title": "Record Session",
+        "desc": "App will automatically record data while in Dashboard mode and will be available as a saved session for immediate analysis",
```
ryandoherty: Suggested wording:

Log Data
Automatically record data in app while in Dashboard mode. Data will be available as a saved session in the Analysis screen.

autosportlabs/racecapture/settings/prefs.py
brentpicasso: Actually because booleans are incorrect for how preferences work. The value stored is actually a string; when I set it to True, the preferences screen did not interpret it correctly. True needs to be represented as '1'.  To be consistent I fixed the telemetry default as well.

resource/settings/settings.json
brentpicasso: How about: 

"Automatically record data in app while in Dashboard mode. Data will be available as a saved session in Analysis mode."

So we consistently use the word 'mode'

autosportlabs/racecapture/settings/prefs.py
ryandoherty: Ok. Huh, then how did this work before? Typecasting?

brentpicasso: Yeah. it typecasted False to '0' but it won't typecast True to '1' . Since telemetry default was 'False' we never noticed it until now.

PR #1236, opened 2016-09-14
Fix crash when datastore has session/lap with inadequate data

ryandoherty:
It's not pretty, but it works. I'm not sure how else to handle the data other than just ignoring data like this completely?

Issue #1230, opened 2016-09-14
Error adding session


brentpicasso:
Another case where I think the session time was super short. Trace:

 File "/home/brent/git-projects/RaceCapture_App/autosportlabs/racecapture/views/analysis/addstreamview.py", line 216, in add_session
    self.dispatch('on_add')
  File "kivy/_event.pyx", line 714, in kivy._event.EventDispatcher.dispatch (/tmp/pip-build-9aBK9m/kivy/kivy/_event.c:7271)
  File "kivy/_event.pyx", line 1224, in kivy._event.EventObservers.dispatch (/tmp/pip-build-9aBK9m/kivy/kivy/_event.c:12786)
  File "kivy/_event.pyx", line 1148, in kivy._event.EventObservers._dispatch (/tmp/pip-build-9aBK9m/kivy/kivy/_event.c:12363)
  File "/home/brent/git-projects/RaceCapture_App/autosportlabs/racecapture/views/analysis/addstreamview.py", line 180, in add_session
    self.dispatch('on_add', list_item.session)
  File "kivy/_event.pyx", line 714, in kivy._event.EventDispatcher.dispatch (/tmp/pip-build-9aBK9m/kivy/kivy/_event.c:7271)
  File "kivy/_event.pyx", line 1224, in kivy._event.EventObservers.dispatch (/tmp/pip-build-9aBK9m/kivy/kivy/_event.c:12786)
  File "kivy/_event.pyx", line 1148, in kivy._event.EventObservers._dispatch (/tmp/pip-build-9aBK9m/kivy/kivy/_event.c:12363)
  File "/home/brent/git-projects/RaceCapture_App/autosportlabs/racecapture/views/analysis/addstreamview.py", line 65, in add_session
    self.dispatch('on_add_session', session)
  File "kivy/_event.pyx", line 714, in kivy._event.EventDispatcher.dispatch (/tmp/pip-build-9aBK9m/kivy/kivy/_event.c:7271)
  File "kivy/_event.pyx", line 1224, in kivy._event.EventObservers.dispatch (/tmp/pip-build-9aBK9m/kivy/kivy/_event.c:12786)
  File "kivy/_event.pyx", line 1148, in kivy._event.EventObservers._dispatch (/tmp/pip-build-9aBK9m/kivy/kivy/_event.c:12363)
[...]

PR #1243, opened 2016-09-15
P/brent/lap data broken

brentpicasso:
I did a poor job reviewing #1236 and jumped the gun approving it. 

The original problem in #1230 is really limited to cases where a single session was recorded - like static testing with no laps.  

Using spaces for lap IDs was causing heartburn for other parts of the analysis view. The proposed change in this PR makes the lap '0'  so the rest of the analysis views work as expected. This is safe since only one session or 'lap' would be generated for this type of session recording, anyway.

The original changeset was revered and an addition was made to datastore.py, noted below. 

@ryandoherty  sorry for my sloppy review.

### DIFF
--- a/autosportlabs/racecapture/datastore/datastore.py
+++ b/autosportlabs/racecapture/datastore/datastore.py
@@ -996,7 +996,9 @@ class DataStore(object):
                                 GROUP BY LapCount, session_id
                                 ORDER BY datapoint.LapCount ASC;''',
                                 (session_id,)):
-            laps.append(Lap(session_id=row[0], lap=row[1] - 1, lap_time=row[2]))
+            lap = row[1]
+            lap = 0 if lap is None else lap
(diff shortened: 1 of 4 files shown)


autosportlabs/racecapture/datastore/datastore.py
```diff
@@ -996,11 +996,9 @@ def get_laps(self, session_id):
                                 GROUP BY LapCount, session_id
                                 ORDER BY datapoint.LapCount ASC;''',
                                 (session_id,)):
-
-            # It's possible a lap has no lap count
-            lap = row[1] - 1 if row[1] else None
-
-            laps.append(Lap(session_id=row[0], lap=lap, lap_time=row[2]))
+            lap = row[1]
```
brentpicasso: this is the actual change made after the revert.

ryandoherty:
Looks good, :shipit: !

PR #1246, opened 2016-09-15
smarter and friendlier time vs. distance selections

### LINKED ISSUE #1245: Need smarter time-vs-distance graphs.
If the user selects plot by distance and no distance data is present, we should be helpful.

The app should auto select plot by time if loading a lap for the first time, but if a user manually switches to distance and no to minimal distance is present, display a warning to explain the empty graph. 

Also, the threshold should be reasonable so some GPS jitter doesn't cause a false positive distance value.

brentpicasso:
Resolves #1245

Also fixes minor bug rendering sessionlistview when there are no sessions.


ryandoherty:
Nice! Also fixes the problem I encountered earlier today. :shipit:

PR #1248, opened 2016-09-15
resize accordion correctly when sessions are added and selected

### LINKED ISSUE #1247: Accordion not resizing correctly after loading sessions
Sometimes when a number of sessions are selected the available space isn't computed correctly, causing an exception, and UI corruption results.

brentpicasso:
Resolves #1247

### DIFF
--- a/autosportlabs/racecapture/views/analysis/analysisview.py
+++ b/autosportlabs/racecapture/views/analysis/analysisview.py
@@ -251,7 +251,6 @@ class AnalysisView(Screen):
 
     def on_add_session(self, instance, session):
         Logger.info("AnalysisView: on_add_session: {}".format(session))
-        self._datastore.refresh_session_data()
         self.check_load_suggested_lap(session.session_id)
         self.ids.sessions_view.append_session(session)
         self.check_load_suggested_lap(session.session_id)
--- a/autosportlabs/racecapture/views/analysis/analysisview.py
+++ b/autosportlabs/racecapture/views/analysis/analysisview.py
@@ -251,6 +251,7 @@ class AnalysisView(Screen):
 
     def on_add_session(self, instance, session):
         Logger.info("AnalysisView: on_add_session: {}".format(session))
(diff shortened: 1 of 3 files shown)


autosportlabs/racecapture/views/analysis/analysisview.py
```diff
@@ -251,6 +251,7 @@ def close_popup(self, *args):
 
     def on_add_session(self, instance, session):
         Logger.info("AnalysisView: on_add_session: {}".format(session))
+        self._datastore.refresh_session_data()
```
ryandoherty: Why does the analysis view have to tell the datastore to refresh its session data?

brentpicasso: it's really the CachingAnalysisDatastore it's talking to.

ryandoherty: Either way, why does it tell it to refresh?

brentpicasso: Ah. so - because we've just imported a new session (on_add_session) and we need to purge the cache.

ryandoherty: Ok. It doesn't seem quite right to me for the analysis view to manage the state of the caching datastore's cache. I'm wondering if the caching datastore should update its internal cache when things request sessions or laps?

brentpicasso: Maybe it could invalidate it automatically when a session is imported or deleted?

ryandoherty: That sounds like a nice way to do it 👍

brentpicasso: Ok, updated to have CachingDatastore handle it internally / automatically.

ryandoherty:
:shipit:

PR #1249, opened 2016-09-16
ensure lap ids are represented by key vs ordinal. 

brentpicasso:
Problem was exposed when importing a dataset that had lap IDs that didn't start at 0. 

has code that depends on #1248, so please review after that one is complete and merged into this branch.


ryandoherty:
:shipit:

Issue #1250, opened 2016-09-16
session recorder perferences regression - stuck in low telemetry rate in dash mode


brentpicasso:
Found that when session recording is disabled, telemetry rate is stuck at 1Hz, even if you enable dashboard mode.  

Need to set telemetry rate if we're recording or if we're in dashboard mode.

PR #1251, opened 2016-09-16
enable active telemetry mode if we're in dashboard mode 

brentpicasso:
resolves regression identified in #1250


ryandoherty:
:shipit:

Issue #1257, opened 2016-09-19
Have exit app button in left popout menu (for desktop only)


brentpicasso:
Include an exit menu option for the left menu options drawer, available on desktop only. 

Suggested fontawesome icon: fa-eject [&#xf052;]

Not a lot of good matches for icons, open to suggestions

ddimensia:
I was thinking of using the fa power-off icon but eject would work also.

Is there a reason not to allow it show it on mobile?  The obvious concern I would have is that it could be accidentally hit via a touch screen.  My plan was to place it at the bottom of the main menu so there was space between it and the System Status item. 

Example:
<img width="1027" alt="screen shot 2016-09-20 at 11 11 00 pm" src="https://cloud.githubusercontent.com/assets/1389554/18700153/8e605e0c-7f8a-11e6-81e2-7a0053ffe153.png">

ddimensia:
I have the changes made up against the 1.7.5 branch.  There's no 1.7.6 branch created yet so I can't send up a PR.  Currently, I have it written so that it removes the 'Exit' menu item if the platform is IOS or Android.  That bit of code could easily be removed if it was determined to be acceptable on any platform.

brentpicasso:
r/1.7.6 branch is up now.  I missed seeing the power button; I like it. 

No reason to not show it on all platforms, actually!

ddimensia:
Still don't see a r/1.7.6 branch, did you push to origin/github?

brentpicasso:
Ok, just pushed!

brentpicasso:
Thanks @ddimensia  for the nice enhancement!

PR #1260, opened 2016-09-21
Resolves autosportlabs/RaceCapture_App#1257, Adds 'Exit' item to left menu.

ddimensia:
This adds an 'Exit' menu item to the bottom of the left hand menu for all platforms.  Currently it uses app.stop() to close the application.

### DIFF
--- a/autosportlabs/racecapture/menu/mainmenu.py
+++ b/autosportlabs/racecapture/menu/mainmenu.py
@@ -32,36 +32,45 @@ MAIN_MENU_KV = """
         Rectangle:
             pos: self.pos
             size: self.size
-    GridLayout:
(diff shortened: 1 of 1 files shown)


ryandoherty:
LGTM!

ddimensia:
The only thing I didn't like in my testing (only on OSX run via python main.py) is that there is a bit of a delay between when you press the menu item and the window closes.  I'm thinking maybe the better thing is to call window.close on the current window prior to app.stop.  Might try that this evening.

ddimensia:
Played with different technics to close the window earlier, all result in a crash with kivy so I think the current implementation is the best I can come up with.

PR #1262, opened 2016-09-22
Fix disconnect when socket command queue is empty

### LINKED ISSUE #1210: Connection drops if CPU is excessively loaded
Noticed that the current connection will drop if the CPU is excessively loaded. Observed with WiFi SocketConnection. 

maybe extend timeout?

ryandoherty:
So this is a bug or copy/pasta typo where the command thread for the socket connection disconnects if no messages are in the command queue. It should just keep running if no messages are there, not disconnect. Seems to fix #1210


brentpicasso:
I can't get it to drop the connection after this fix. 

I say :shipit:  but only if you take back all the mean things you said about python :P :)

ryandoherty:
Hahahhahahahah

PR #1263, opened 2016-09-22
ensure old received data is blanked out when socket closes

### LINKED ISSUE #1189: Loss of WiFi connection requires re-start of app
Observed, need to reproduce: When the TCP connection drops, app sometimes gets into a mode where it cannot re-connect; requiring a re-start of the app.

This issue is to track repro and resolution.

brentpicasso:
resolves #1189


ryandoherty:
:shipit: !!!

Issue #1077, opened 2016-07-19
Hide firmware update menu item if RC device is connected over wifi


ryandoherty:

ddimensia:
Seems like the best way to do this might be to add a method to the RcpApi class that determines if a feature is supported.  In this case the feature would be firmware update.  The RcpApi could either defer that to the comms object or return a value based on the type of the comms object.  This would also mean the current FIRMWARE_UPDATABLE property in configview.py should probably be removed in preference of the new RcpApi method.

Neither the SystemSettings nor RcpConfig classes seem appropriate.

brentpicasso:
This is a function of the interface - where firmware updates are only supported over USB - so having it served up via RcpApi through the specific connection as a type of feature flag or boolean class property seems appropriate.

PR #1267, opened 2016-10-04
Issue #1077: Hide Firware update menu item when using wifi.

ddimensia:
This may have to be revisited if the app ever supports switching connection types at runtime but for now it seems to work as intended.

### DIFF
--- a/autosportlabs/racecapture/api/rcpapi.py
+++ b/autosportlabs/racecapture/api/rcpapi.py
@@ -758,6 +758,9 @@ class RcpApi:
         else:
             self.sendCommand({'s':0})
 
+    def is_firmware_update_supported(self):
+        return self.comms and not self.comms.is_wireless()
+
     def start_auto_detect_worker(self):
         self._auto_detect_event.clear()
         t = Thread(target=self.auto_detect_worker)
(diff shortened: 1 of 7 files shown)


autosportlabs/racecapture/api/rcpapi.py
```diff
@@ -758,6 +760,9 @@ def sample(self, include_meta=False):
         else:
             self.sendCommand({'s':0})
 
+    def is_firmware_update_supported(self):
+        return not is_mobile_platform() and not isinstance(self.comms, SocketComm)
```
brentpicasso: IMO it would be good to not have rcpapi have any knowledge of specific Comms objects. I wonder if we could push that knowledge down to the individual Comms objects, or put the knowledge in the comms_factory. 

firmware upgrading can only happen on a 'direct connection' (usb). or a 'non wireless' connection. perhaps we can have an is_wireless property on each comms object and derive if firmware is update-able based on whether the connection is not wireless?

ddimensia:
Rewrote to rely on is_wireless method on Comms objects, dropped is_mobile_platform check as that was just a surrogate for the is_wireless check.   Updated unit tests and verified manually via OSX + RaceCapture unit.

brentpicasso:
Looks good. merging!

PR #1269, opened 2016-10-05
Replace 'race-capture.com' with 'Podium' in preferences resource file.

### LINKED ISSUE #1268: Preferences screen references race-capture.com under Send Telemetry
I noticed that the app still makes reference to race-capture.com in the preferences screen.  It should instead reference 'Podium'.

ddimensia:
resolves #1268


brentpicasso:
Rabble! I thought we got all of those. :)

PR #1271, opened 2016-10-08
Issue #1195: Delay session stop after dashboard is exited by one minute.

### LINKED ISSUE #1195: Delay session stop to allow accidental exit / re-enter of dashboard mode.
If the user exits dashboard mode by accident, the session should not stop immediately. Session could stop some time (say one minute) after dashboard is exited. 

This would prevent excessive number of small sessions.

ddimensia:
Resolves #1195 

I added a timer that gets started on when the view leaves the dashboard.  If the view returns to the dashboard then the timer will be canceled, if the timer fires then the session will actually be stopped.  The default delay is 60 seconds, currently this isn't configurable in the app but could probably be added to the app preferences.

### DIFF
--- a/autosportlabs/racecapture/data/sessionrecorder.py
+++ b/autosportlabs/racecapture/data/sessionrecorder.py
@@ -18,11 +18,11 @@
 # have received a copy of the GNU General Public License along with
 # this code. If not, see <http://www.gnu.org/licenses/>.
 
-from datetime import datetime
-from autosportlabs.racecapture.datastore import Session
+from kivy.clock import Clock
 from kivy.logger import Logger
 from kivy.event import EventDispatcher
 from autosportlabs.util.timeutil import format_time
+
 
 class SessionRecorder(EventDispatcher) :
     """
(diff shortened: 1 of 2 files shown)


ddimensia:
I need to update/add tests and I think making this configurable may be a good idea although that could be added as a separate Issue/PR if you'd like to get this in sooner.

ddimensia:
Updated and added tests, fixed an issue where disconnect didn't stop the session immediately.  I also made it so that the stop timer can be disabled if the stop_delay is set to <= 0.  I squashed my commit into one as to keep the commit history reasonably clean.

brentpicasso:
awesome - will check it out this weekend!

brentpicasso:
Recommend using Kivy's Clock.create_trigger mechanism instead of creating a thread; this keep the processing in the UI thread. 

For a great example, see toolbarview.py - 

line 286, 299 and 218 for an example of a trigger that can be cancelled.

ddimensia:
Updated to use Clock.create_trigger instead of Timer.  Update unit tests and verified changes locally via OSX + RC.

brentpicasso:
tested, works great. :ship:

PR #1272, opened 2016-10-10
Mark RCPAPI worker threads as daemon threads, clean up some deprecated calls.

ddimensia:
This PR has two commits, the first just fixes some deprecated call warnings that were showing up in the logs and should have no effect on how the app behaves.

The second commit is to resolve a few issues I've noticed while running the app using Linux on a Raspberry Pi 3.  The issues also present themselves on OSX but are not as big of a deal given that the application isn't fullscreen like it is on Linux.  

There are basically two issues that manifest in the same behavior.  First, if a user tries to kill the app using a SIGINT or CTRL-C then kivy tries to kill the application but the background worker threads keep the application from dying.  Second, if an unhandled exception is thrown in the application there is the appears that the application freezes, however, kivy is trying to close the app, similar to a SIGINT, but stays active due to the worker threads.

By switching the background threads to be daemon threads it allows the kivy app to close if either a SIGINT is received or if an unhandled exception occurs.

### DIFF
--- a/autosportlabs/racecapture/api/rcpapi.py
+++ b/autosportlabs/racecapture/api/rcpapi.py
@@ -133,6 +133,7 @@ class RcpApi:
     def _start_message_rx_worker(self):
         self._running.set()
         t = Thread(target=self.msg_rx_worker)
+        t.daemon = True
         t.start()
         self._msg_rx_thread = t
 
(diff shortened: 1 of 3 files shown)


brentpicasso:
Looks good. thanks for the fixes!

PR #1279, opened 2016-10-17
Issue #1065: Added license header to all .py, .kv, .java, and .sh files where appropriate

ddimensia:
I also fixed some spacing issues in some files that already head the header.


brentpicasso:
Wow. thanks!

Issue #931, opened 2016-04-11
IndexError: list index out of range


stieg:
Got this when importing a fixed log file where the names had spaces (Issue #901).  Don't know the cause but we should handle it...

```
[INFO              ] [DataStore   ] Created session with ID: 5
 Exception in thread Thread-11:
 Traceback (most recent call last):
   File "/usr/lib64/python2.7/threading.py", line 813, in __bootstrap_inner
     self.run()
   File "/usr/lib64/python2.7/threading.py", line 766, in run
     self.__target(*self.__args, **self.__kwargs)
   File "/home/stieg/Devel/asl/rc_app/autosportlabs/racecapture/views/analysis/addstreamview.py", line 183, in _loader_thread
     session_id = self.datastore.import_datalog(logpath, session_name, session_notes, self._update_progress)
   File "/home/stieg/Devel/asl/rc_app/autosportlabs/racecapture/datastore/datastore.py", line 23, in wrap
     ret = f(*args)
   File "/home/stieg/Devel/asl/rc_app/autosportlabs/racecapture/datastore/datastore.py", line 797, in import_datalog
     self._handle_data(dl, headers, session_id, warnings, progress_cb)
   File "/home/stieg/Devel/asl/rc_app/autosportlabs/racecapture/datastore/datastore.py", line 681, in _handle_data
     cur.executemany(datapoint_sql, datapoint_iter(newdata_gen, starting_datalog_id))
   File "/home/stieg/Devel/asl/rc_app/autosportlabs/racecapture/datastore/datastore.py", line 662, in datapoint_iter
     for record in data:
   File "/home/stieg/Devel/asl/rc_app/autosportlabs/racecapture/datastore/datastore.py", line 564, in _desparsified_data_generator
     work_list[c].append(channels[c])
 IndexError: list index out of range
 ```

[RC_8_fixed.LOG.txt](https://github.com/autosportlabs/RaceCapture_App/files/213595/RC_8_fixed.LOG.txt)

stieg:
We need to be able to discard lines if they are improperly formatted for whatever reason.

ddimensia:
I believe this was fixed with 5aba116dc53ed5859f6b5981a6839439380d13fe, 92539e6e9740435fa9d84f00fe280853bf059017 and  8ee2298090e5cbc189d0e7a45ed0690ba6ae7872.  Within those commits the log file is checked to make sure that the channel list is less than the number of elements in the log.

[...]

PR #1280, opened 2016-10-17
Issue #931: Fixed issue with line number not being incremented

ddimensia:
The 'current_line' wasn't being incremented which would allow the code to fail if the second line was malformed.  

I moved the current_line += 1 to the bottom of the loop.  The only effect this would have verses it's previous placement would be on the calculation of the percentage status bar.  It could be considered off by one but given that the dialog disappears at the end of the loading I don't know if would be noticeable by the end user.


brentpicasso:
tested, seems legit. :ship:

PR #1291, opened 2016-11-03
added GPS warning on higher GPS sample rates

### LINKED ISSUE #956: Create Warning about Setting GPS to 50Hz
Too many customers are setting their GPS to 50Hz without understanding the requirements.  We need a warning in the App that explicitly tells them this is a bad idea unless they are in the middle of a desert (or Button willow... which is effectively the same thing).  Hopefully this will cut down on the number of times we have to clarify this.

brentpicasso:
* Added GPS warning message on sample rates >= 25Hz
* Added help popup with extended information
* Fixed minor bug where help popup would not show in correct location on phone form factor
* Fixed minor bug for analog scaling map where warning bubble would cover text box
* kv file refactoring (moved kv into source)
* added drop-shadow around help popup to help it stand out from background
resolves #956


brentpicasso:
image

ryandoherty:
Looks good! I could only get the modal to pop up once, couldn't tell if that was intentional or not from the code.

brentpicasso:
Correct - the help info screens are designed to pop up only once, to show or advise things when people are learning the system for the first time.

Issue #1285, opened 2016-10-24
SocketComms data can fall behind and never catch up


brentpicasso:
the current way socketComms is currently designed is that the queue will acculmulate indefinitely. The consumer always runs at least at the same rate or lower.  If the main thread is backed up, it will progressively fall behind and never recover. 

We need to limit how much data is queued so data is dropped if needed to keep the UI responsive.

stieg:
Would it instead just be wiser to drain the queue on every iteration of the refresh instead of taking a single sample.  This way everything gets processed but only the latest update shows up on the UI.

brentpicasso:
possibly. the main reason for this falling behind is due to slower devices, so we will have to drop data at some point.

stieg:
> the main reason for this falling behind is due to slower devices

The device producing this data has trivial resources relative to even the oldest ones supported (Old galaxy tabs or Kindle fire 7") that receive and process this data and you are telling me that these more powerful devices can't keep up?  From where I sit this would indicate a larger design issue as the culprit rather than the devices themselves.

PR #1293, opened 2016-11-04
reduce receive buffer in SocketConnection to prevent wifi lag

brentpicasso:
So it seems that various network and system hiccups would cause API messages to stack up inside the original 4096 buffer we defined in SocketConnection, eventually causing a visual lag. 

Reduced buffer to 32 so that we process chunks in finer grained increments.  After this change performance and responsiveness seems great, regardless of direct connection or via intermediary access point.

Thinking of making buffer size 64; 32 guarantees we always quickly process smaller messages

### DIFF
--- a/autosportlabs/comms/socket/socketconnection.py
+++ b/autosportlabs/comms/socket/socketconnection.py
@@ -23,7 +23,6 @@ import socket
 import json
 import errno
 
-PORT = 7223
 READ_TIMEOUT = 2
 SCAN_TIMEOUT = 3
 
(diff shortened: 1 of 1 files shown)


brentpicasso:
resolves #1285

ryandoherty:
This seems ok, but I'm not able to see any difference in lag for 32 vs 4096. Is it worse on slower platforms?

brentpicasso:
The lag can come and go.  Try this:

with 4096 have wifi connected via desktop app wifi connection.

configure main gauge with yaw sensor.
observe low latency.
put the app in the background, but frozen (ctrl-z)
wait a sec or so. 
put the app back in the foreground.  
now move racecapture around and observe delay.

repeat with this PR code and observe it never backs up.

brentpicasso:
It seems to happen more on slower platform or slower / congested networks, yes

ryandoherty:
I'm unable to repro, but I believe you when you say you can see it. So long as it's tested on Android and works then :shipit:

brentpicasso:
Okie.

PR #1295, opened 2016-11-04
warning help when changing WiFi settings on mobile

### LINKED ISSUE #1202: Consider having thea App warn users about changing WiFi settings when connecting using Wifi with device
Doing so could cause a disruption and failure when writing  a config since the Wifi will apply the changes as soon as it gets them.  Currently I will only apply new changes if the configuration changed and I wait 1 second to do this so as to give the reply message time to send.

brentpicasso:
Looks like this. Fixes #1202 

image


autosportlabs/racecapture/views/configuration/rcp/wireless/wificonfigview.py
```diff
@@ -128,68 +131,73 @@ def __init__(self, base_dir, config, **kwargs):
 
         self.register_event_type('on_modified')
 
+    def _wifi_modified(self):
+        if is_mobile_platform():
+            Clock.schedule_once(lambda dt: HelpInfo.help_popup('wifi_warning', self.parent.parent), 1.0)
```
ryandoherty: Shouldn't this occur if the connection is currently over Wifi?

brentpicasso: that would be ideal, but I think it's ok to remind people anyway even if they're connected over BT. 

The ROI of wiring up detecting if we're currently on WiFi is not worth it IMO.

ryandoherty: Ok 👍

ryandoherty:
:shipit:

PR #1298, opened 2016-11-09
unique session name based on venue name first, then a date stamp

### LINKED ISSUE #1256: make autogenerated session name more concise
Currently it makes a <date> <time> format, which is very long.

Suggested format:

<date> - <ordinal>

like:

1/1/2001 - 1
1/1/2001 - 2

This will fit nicely into various areas of the analysis view.

brentpicasso:
resolves #1256

### DIFF
--- a/autosportlabs/racecapture/data/sessionrecorder.py
+++ b/autosportlabs/racecapture/data/sessionrecorder.py
@@ -21,8 +21,9 @@
 from kivy.clock import Clock
 from kivy.logger import Logger
 from kivy.event import EventDispatcher
-from autosportlabs.util.timeutil import format_time
-
+from autosportlabs.util.timeutil import format_date
+from autosportlabs.racecapture.config.rcpconfig import GpsSample
+from autosportlabs.racecapture.geo.geopoint import GeoPoint
 
 class SessionRecorder(EventDispatcher) :
     """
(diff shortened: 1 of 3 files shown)


autosportlabs/racecapture/data/sessionrecorder.py
```diff
@@ -138,11 +148,25 @@ def _check_should_record(self):
...
     def _create_session_name(self):
         """
-        Creates a session name
+        Creates a session name attempting to use the nearby venue name first, 
+        then falling to a date stamp
         :return: String name
         """
-        date_string = format_time()
-        return date_string
+        nearby_venues = self._track_manager.find_nearby_tracks(GeoPoint.fromPoint(self._gps_sample.latitude, self._gps_sample.longitude))
```
ryandoherty: This doesn't seem to account for if a user selects a track manually for RCT or if someone has put 1 track on RCP out of a possible N tracks. I think this would end up with sessions with the wrong or missing track name.

brentpicasso: And they can go rename the session if they want to, as well.

brentpicasso: If RC / RCP has already been plugged in before the app launches, then there's a good chance it already has a GPS lock, so that concern is pretty low, IMO.

ryandoherty:
:shipit:

PR #1299, opened 2016-11-09
save current analysis channels in preferences

### LINKED ISSUE #1096: Analysis view should save settings
When configuring the analysis view's channels, graph by (time/distance) and any other user configuration, it should be saved so the next time the view is loaded, it is configured the same way it was before.

brentpicasso:
resolves #1096

### DIFF
--- a/autosportlabs/racecapture/settings/prefs.py
+++ b/autosportlabs/racecapture/settings/prefs.py
@@ -375,8 +375,10 @@ class UserPrefs(EventDispatcher):
         :param value the preference value to set
         :type value list (list of strings)
         """
-        self.set_pref(section, option, ','.join(value))
-        
+        try:
+            self.set_pref(section, option, ','.join(value))
+        except TypeError:
+            Logger.error('UserPrefs: failed to set preference list for {}:{} - {}'.format(section, option, value))
         
     def to_json(self):
         '''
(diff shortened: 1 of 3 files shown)


autosportlabs/racecapture/settings/prefs.py
```diff
@@ -346,6 +365,19 @@ def set_pref(self, section, option, value):
...
+        """
+        Set a preference value by list
+        :param section the configuration section for the preference
+        :type string
+        :param option the option for the section
+        :type string
+        :param value the preference value to set
+        :type value list (list of strings)
+        """
+        self.set_pref(section, option, ','.join(value))
```
ryandoherty: What happens here if `value` isn't an object/type that `join()` accepts? Might be worth it to add some error catching if necessary.

brentpicasso: I could catch, log and re-throw so crash handler gets it, or, I could catch, log, and swallow the error. 

        try:
            self.set_pref(section, option, ','.join(value))
        except TypeError:
            Logger.error('UserPrefs: failed to set preference list for {}:{} - {}'.format(section, option, value))


Thoughts on one approach vs the other? Saving a preference isn't critical for operation; it's a convenience for the user. If we re-throw it could disrupt the function attempting to run.  if we swallow it, we won't find out about it via CrashHandler.

Have we been using CrashHander in a soft way yet?

brentpicasso: Updated to log and swallow the exception. Can change it based on discussion.

ryandoherty: Logging and swallowing exception sounds fine to me since the end result will just be the app not remembering preferences. We'll then have logs to help us deduce the problem 👍

ryandoherty:
:shipit:

Issue #967, opened 2016-05-02
Customize what screens are available in dashboard mode


brentpicasso:
Allow selection of what screens are available to cycle through in dashboard mode. Likely enabled through dashboard configuration screen.

PR #1315, opened 2016-12-16
P/brent/nextdash

brentpicasso:
Add more dashboard screens and make them selectable.


brentpicasso:
resolves #967

autosportlabs/racecapture/views/dashboard/dashboardview.py
```diff
@@ -45,18 +49,106 @@
...
+        self._init_factory()
+
+    def create_screen(self, key):
+        """
+        Create a new dashboard screen instance by key
+        :param key the key of the screen
+        :type key String
+        :return a new instance of a dashboard screen
+        """
+        view = self._view_builders[key]()
```
ddimensia: This should probably handle an invalid key more gracefully than throwing a KeyError.  The potential issue is if a key gets renamed or removed, this would make the preferences invalid.  I guess that could be handled at loading time in the DashboardView class.

autosportlabs/racecapture/views/dashboard/dashboardview.py
```diff
@@ -45,18 +49,106 @@
...
+        """
+        Get the preview image source path for the specified dashboard key
+        :param key the key of the dashboard
+        :param key String
+        :return String path to the preview image, or None if the preview is not found
+        """
+        return self._view_previews.get(key)
+
+    def build_5x_gauge_view(self):
+        return GaugeView5x(name='5x_gauge_view', databus=self._databus, settings=self._settings)
```
ddimensia: It'd be nice if the keys/names could be defined once to reduce the risk of a copy/paste or typo causing a problem.  Maybe have the builder accept the name?  The reason I bring this up is because the last_dash_screen pref is based on the 'name' of the view.

autosportlabs/racecapture/views/dashboard/dashboardview.py
```diff
@@ -200,7 +267,12 @@ def _init_view(self):
         settings = self._settings
 
         self._init_global_gauges()
-        self._init_view_builders()
+
+        # add the initial set of empty screens
+        screens = self._screens
+        screens += self._settings.userPrefs.get_dashboard_screens()
```
ddimensia: Should probably validate that the screens in prefs are still valid.  If this filtered invalid screens out then show_last would still work if it referenced an invalid screen since it won't be in the 'screens' list.

autosportlabs/racecapture/views/dashboard/dashboardview.py
brentpicasso: Agreed in principal.  What I really want is a way for the factory to auto-discover the available dashboard classes through some type of reflection (e.g. find all classes that implement DashboardScreen) - then none of these dashboard screens would be hardcoded.

The dashboard screens themselves would provide the name, and then the caller could request them.

autosportlabs/racecapture/views/dashboard/dashboardview.py
```diff
@@ -45,18 +49,106 @@
 from autosportlabs.racecapture.config.rcpconfig import Track, TrackConfig, Capabilities, GpsConfig, GpsSample
 from autosportlabs.racecapture.tracks.trackmanager import TrackMap, TrackManager
 # Dashboard screens
-from autosportlabs.racecapture.views.dashboard.gaugeview import GaugeView
+from autosportlabs.racecapture.views.dashboard.gaugeview import *
```
brentpicasso: :+1: fixed.

ryandoherty:
Code looks sane overall, the 3 gauge view has the gauges overlapping, which makes my left eye twitch :D

autosportlabs/racecapture/views/dashboard/dashboardview.py
ryandoherty: I agree, if someone upgrades or downgrades the app the key could be missing.

autosportlabs/racecapture/views/dashboard/dashboardview.py
```diff
@@ -44,19 +48,110 @@
...
+
+    @property
+    def available_dashboards(self):
+        """
+        Provide a list of available dashboards by key
+        :return String
+        """
+        return self._view_builders.keys()
+
+    def get_dashboard_preview(self, key):
```
ryandoherty: It would be nice for this function name to denote that the preview is an image or path.

autosportlabs/racecapture/views/dashboard/dashboardview.py
```diff
@@ -220,14 +295,67 @@ def _init_view(self):
...
+
+        # clear all of the dashboard screens from the outer container
+        for screen in loaded_screens.values():
+            parent = screen.parent
+            if parent is not None:
+                parent.remove_widget(screen)
+
+        # add more carousel panes as needed
+        while True:
+            if current_screen_count == new_screen_count:
```
ryandoherty: What happens if the current screens are not the same as the new screens, but are the same length? Would this not update?

autosportlabs/racecapture/views/dashboard/dashboardview.py
```diff
@@ -359,20 +490,32 @@ def on_key_down(self, window, key, *args):
             self.on_nav_right()
         return False
 
+    def _filter_dashboard_screens(self, screens):
+        """
+        Filter the supplied list of screens against the actual list of available screens, including normalizing for order
+        """
+        all_screens = self._dashboard_factory.available_dashboards
+        # re-order selected screens based on native order
+        return [x for x in all_screens if x in screens]
```
ryandoherty: List comprehension, very pythonic 👍

autosportlabs/racecapture/views/dashboard/dashboardview.py
```diff
@@ -393,30 +536,185 @@ def on_nav_right(self):
...
+    """
+    Builder.load_string(DASHBOARD_SCREEN_PREFERENCES_KV)
+
+    def __init__(self, settings, dashboard_factory, **kwargs):
+        super(DashboardScreenPreferences, self).__init__(**kwargs)
+        self._settings = settings
+
+        current_screens = self._settings.userPrefs.get_dashboard_screens()
+        screen_keys = dashboard_factory.available_dashboards
+        for key in screen_keys:
```
ryandoherty: This block feels long, maybe it could be a separate function?

autosportlabs/racecapture/views/dashboard/dashboardview.py
```diff
@@ -393,30 +536,185 @@ def on_nav_right(self):
...
+            preview = dashboard_factory.get_dashboard_preview(key)
+            name = preview[0]
+            image = preview[1]
```
ryandoherty: Could do list dereferencing to make this shorter:

`[name, image] = dashboard_factory.get_dashboard_preview(key)`

autosportlabs/racecapture/views/dashboard/dashboardview.py
brentpicasso: :)

autosportlabs/racecapture/views/dashboard/dashboardview.py
brentpicasso: I think it's ok; the function is only 25 lines long and is highly readable about what it does.

autosportlabs/racecapture/views/dashboard/dashboardview.py
brentpicasso: I like it, fixing.

autosportlabs/racecapture/views/dashboard/dashboardview.py
brentpicasso: agreed, refining function name.

autosportlabs/racecapture/views/dashboard/dashboardview.py
brentpicasso: This is handled in two places by the use of _filter_dashboard_screens() - in init_view and when preferences are adjusted by the user upon the close of the dialog.  if we hit the key error then the crash handler will catch it and provide notifications accordingly.

PR #1343, opened 2017-03-27
Update canchannelsview.py

jlwall:
Endianess was backwards in the commit, false (0) is little, true (1) is big, the kivy view should represent the same thing.
Tested this with 2.11 with an external can message.


brentpicasso:
Look at that. Thanks for the catch!

Issue #984, opened 2016-05-15
OBD-II - ability to specify a custom PID


brentpicasso:
Currently the list of PIDs are fixed in the app. Need the ability to specify a custom PID and corresponding conversion formula for the data.

Issue #982, opened 2016-05-15
Direct CAN mapping configuration


brentpicasso:
Enable configuration management for mapping CAN data to channels.  Provide an interface that allows defining how to extract data from a CAN message, transform it using a formula, and set it to a channel.

Each Mapping has the following configuration:
- A standard ChannelConfig:
- Name
- Units
- min
- max
- logging precision
- CAN Mapping
- Source data type(unsigned, signed, IEEE754)
- Endian
- CAN channel
- CAN ID
- Bit offset within CAN message
- bit width
- Value multiplier (float)
- Value offset (float)

Compliment issue in firmware:
https://github.com/autosportlabs/RaceCapture-Pro_firmware/issues/172

Work currently tracked in https://github.com/autosportlabs/RaceCapture_App/tree/p/brentpicasso/CAN_mapping

![image](https://cloud.githubusercontent.com/assets/3267533/15275613/25afb2be-1a85-11e6-9a71-be43a48b753b.png)

kosenko-max:
Definition of sample rate in Hz looks confusing. If this is mapping of CAN messages without requests - what is the point of sample rate at all? If it is request-response based data then sample rate probably should be in form of priority (1 means request every cycle, 2 means every second cycle, etc).

For request-response based systems only 1 thing is needed in GUI - CAN request message as hex string. Such a pain to do CAN messaging in Lua - it doesn't fit into memory with the rest of the stuff no matter what I do.

brentpicasso:
@kosenko-max the sample rate is how fast we internally sample the current value, independently of how fast the CAN messages are coming in.

Note, this is not for PID style request / replies, but for mapping streaming CAN messages.

kosenko-max:
I understand it's not for it.

But honestly, I don't see much of the difference that it would require
separate GUI. In my particular case 1 request generates large multi-message
answer containing most of the log data. It requires work with different
byte modes and conversion of results by multiplier and offset. Request is
actually memory address and number of bytes to read from it and it's not a
PID. I don't see anything like that coming to OBD. But I can see how adding
request field makes all this a reality.

On Evo ECU - there is a possibility to tap into ECU harness and start
listening stream of CAN messages - it has same format of data as OBD.
Except it's not enough data.

If I had an option to do that in Lua - I would. But unfortunately it's 50kb
RAM.

brentpicasso:
Thanks. We'll work to accommodate what you need into the system overall, using your current script as a guideline.

PR #1345, opened 2017-03-29
P/brent/next obd2 can

brentpicasso:
UI for next gen CAN and OBDII mappings, plus API and related boilerplate. 

Best to test with 2.11.0 firmware (MK3, MK2 with 2.11.0 RC1)


ryandoherty:
Questions:

1. If I select a preset OBDII channel, then change the tab to OBDII pid and change the pid, the channel name stays on the preselected name. Might be best to change it to 'Uknown' if the user changes any of the other values after selecting a preconfigured OBDII channel.
2. (see image below) Tab ordering sometimes reverses?
image
3. Would be easier to type a number instead of scrolling through hundreds of PID numbers:
image

autosportlabs/racecapture/config/rcpconfig.py
```diff
@@ -828,6 +828,60 @@ def toJson(self):
...
 
+class CANMapping(object):
+    TYPE_UNSIGNED = 0
+    TYPE_SIGNED = 1
+    TYPE_FLOAT = 2
+    ID_MASK_DISABLED = 0
+    CONVERSION_FILTER_DISABLED = 0
+
+    def __init__(self, **kwargs):
+        super(CANMapping, self).__init__(**kwargs)
```
brentpicasso: ...I don't think so. I think it was just habit that I included it this way from all of the kivy ui widgets.

autosportlabs/racecapture/views/configuration/rcp/canchannelsview.py
```diff
@@ -0,0 +1,498 @@
...
+        self.ids.can_id.text = '{}'.format(can_mapping.can_id)
+
+        self.ids.can_offset_len.text = u'{}({})'.format(can_mapping.offset, can_mapping.length)
+
+        sign = '-' if can_mapping.adder < 0 else '+'
+        self.ids.can_formula.text = u'\u00D7 {} \u00F7 {} {} {}'.format(can_mapping.multiplier, can_mapping.divider, sign, abs(can_mapping.adder))
+        self._loaded = True
+
+class CANPresetResourceCache(ResourceCache):
+    preset_url = "http://podium.live/api/v1/can_presets"
```
brentpicasso: fixing!  funny, resourcemanager isn't fully wired up. probably better have it as a separate PR anyway.

brentpicasso:
It's a balance on cluttering a view with too much information and separating things out into a logical but related sections. the CAN bus screen currently has the baud rate selectors, but needs to be updated to also include the CAN termination switches as well, so it's going to have quite a bit of information on one screen already.

A lot of this will go to automated setup wizards when the app is first installed where a series of questions and automated setup will guide people through things.

brentpicasso:
Regarding pre-set selection.  Actually, what's being attempted here is if they select an existing SAE standard channel, the rest of the information is automatically populated throughout the rest of the dialog. 

However, there's a distinct case where they may want to add an OEM specific PID. They're going to set the channel name and discretely customize the PID, formula and etc.  I don't think we can symmetrically go and set the channel name to 'unknown' if they change the PID (for example) - because then they might go change the channel name back again, and then it might overwrite their PID changes - a catch 22.

Reflecting on it more, I think we need to pop up a warning or ask the user if they want the selected preset to change the existing settings.  But maybe only if they're editing a current OBDII channel . If they're adding a new channel, then the preset populates the channel settings quietly.

brentpicasso:
I debated the scrolling through PIDs vs typing a number. I do like how it's crisply constrained to the numbers we specify. I think it's pretty quick to scroll through overall: 1 - 255.  I'd like to keep it unless it proves to be cumbersome based on feedback.

autosportlabs/racecapture/config/rcpconfig.py
brentpicasso: fixed.

brentpicasso:
ok @ryandoherty  - I made a change to how presets are applied. 

* if the channel editor is for a brand new channel, we never prompt. 
* if the current channel matches the incoming preset exactly, then we don't prompt to overwrite preset
* If editing an existing channel *and* the channel was previously customized (and no longer matches the preset for the channel matching that name),  we prompt the user if they want to apply the preset- possibly avoiding the possible overwriting of the channel mapping details.

brentpicasso:
Fixed tab ordering, made preset selection more intelligent, avoiding cases where we overwrite user's customizations. 

Added help popups, too.

brentpicasso:
resolves #984 #982

ryandoherty:
Awesome! The tabs are really nice, good work on those 👍

Issue #1333, opened 2017-02-02
WiFi Config page should disable password field if encryption setting is set to 'none'


brentpicasso:
This will cause confusion b/c the wifi module will disable encryption if encryption type is not set, even if a password is set.

Issue #1324, opened 2016-12-24
Add WiFi initialized indicator to Status page


brentpicasso:
In the app we also need an initialized indicator for WiFi so we can determine if there's a problem initializing the module.  We can determine failure to initialize based on the active and initialized settings:

"wifi":{"active":true,"initialized":true,"ap":{"active":true},"client":{"active":false,"connected":false}}

Issue #1283, opened 2016-10-18
WiFi AP password needs input validation


brentpicasso:
AP pass phrase should be a minimum of 8 characters in wifi config

PR #1349, opened 2017-04-03
add wifi AP password validation + enabling tab stops between fields + improved wifi status

### LINKED ISSUE #1011: Pressing tab key doesn't move to next form field
In normal apps and webpage, pressing the Tab key will move your cursor to the next form field. This doesn't happen in our app, due to it being Kivy based. Need to figure out if this is possible.

brentpicasso:
Resolves #1011  #1283  #1333 #1324

### DIFF
--- a/autosportlabs/uix/bettertextinput.py
+++ b/autosportlabs/uix/bettertextinput.py
@@ -20,9 +20,12 @@
 
 import kivy
 kivy.require('1.9.1')
+from kivy.clock import Clock
+from kivy.metrics import sp
+from autosportlabs.racecapture.views.popup.centeredbubble import CenteredBubble, WarnLabel
 from valuefield import ValueField
+from autosportlabs.racecapture.theme.color import ColorScheme
 import re
-
 
 class BetterTextInput(ValueField):
     """
@@ -30,10 +33,15 @@ class BetterTextInput(ValueField):
     entered.
     """
 
+    WARN_SHORT_TIMEOUT = 0.25
+    WARN_LONG_TIMEOUT = 10.0
+
+
     def __init__(self, max_chars=200000, regex=None, **kwargs):
         super(BetterTextInput, self).__init__(**kwargs)
         self.max_chars = max_chars
         self.regex = regex
+        self.warn_bubble = None
 
     def insert_text(self, substring, from_undo=False):
         if not from_undo and (len(self.text) + len(substring) > self.max_chars):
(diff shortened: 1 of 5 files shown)


valuefield.py
```diff
@@ -28,6 +28,7 @@
 <ValueField>:
     font_size: self.height * 0.5
     multiline: False
+    write_tab: False
```
brentpicasso: This makes all uses of ValueField respond to the tab key automatically, moving the focus to the next focusable field, automatically, as expected in a regular app.

autosportlabs/uix/bettertextinput.py
```diff
@@ -44,3 +52,23 @@ def insert_text(self, substring, from_undo=False):
...
+            warn.add_widget(WarnLabel(text=str(error), font_size=sp(12)))
+            warn.background_color = ColorScheme.get_error()
+            warn.auto_dismiss_timeout(self.WARN_LONG_TIMEOUT)
+            warn.size = (self.width, sp(50))
+            warn.size_hint = (None, None)
+            p = self.get_root_window()
+            p.add_widget(warn)
+            warn.center_above(self)
+            self.warn_bubble = warn
+            Clock.schedule_once(lambda dt: self.clear_error(), BetterTextInput.WARN_LONG_TIMEOUT)
```
ryandoherty: Along with hiding it after a timeout, it should also probably hide if the user switches to a new config tab or screen.

autosportlabs/uix/bettertextinput.py
```diff
@@ -44,3 +52,23 @@ def insert_text(self, substring, from_undo=False):
...
         super(BetterTextInput, self).insert_text(substring, from_undo=from_undo)
+
+    def set_error(self, error):
+        if self.warn_bubble is None:
+            warn = CenteredBubble()
+            warn.add_widget(WarnLabel(text=str(error), font_size=sp(12)))
+            warn.background_color = ColorScheme.get_error()
+            warn.auto_dismiss_timeout(self.WARN_LONG_TIMEOUT)
+            warn.size = (self.width, sp(50))
+            warn.size_hint = (None, None)
```
ryandoherty: Seems like we could create a reusable component for warning widgets that center above/below form fields. Not necessary for this PR, but will probably come in handy in the future.

autosportlabs/uix/bettertextinput.py
brentpicasso: aw yeah. We're adding it to the top level window, which sits above all other windows. Maybe it needs to be limited to the current screen / tab / etc. Good catch

brentpicasso:
I gave up on the popup bubble and just went with a fieldLabel to show the error. not as sexy, but simple and reliable. 

image

brentpicasso:
Going ahead and merging this PR - addressed remaining outstanding issue with simplified warning.

Issue #1334, opened 2017-02-03
Error if a saved channel exists in preferences, yet disappears from datastore or session


brentpicasso:
If a channel is remembered, but the loaded session no longer has that channel, an error is displayed. 

STR:
load datastore with a distinct channel. e.g. sonoma with "OilTemp". 
Select that channel for display
Delete datastore 
Launch app, import a different logfile that does not have that channel. 
Observe error:
![image](https://cloud.githubusercontent.com/assets/3267533/22606914/722f3768-ea0a-11e6-9849-a3636b2762db.png)

Issue #1331, opened 2017-02-01
analysis import fails silently on log files when they fail to import


brentpicasso:
If the import fails for any reason the app just hangs on the import screen without any feedback. this is bad :)

Issue #1330, opened 2017-02-01
Analysis fails to import channel starting with number in the name


brentpicasso:
STR: 

Create channel starting with a number.  Import data log. Observe data log import failure in command line. 

Should gracefully handle column names starting with a number.

PR #1350, opened 2017-04-03
P/brent/analysis robusto

### LINKED ISSUE #1319: Exceptions thrown when non-existent channel is selected
STR:

import a datalog with a set of channels;
select a channel to display
remove datalog
import a different datalog with channels not present in the first
select a lap
the cached channel name will attempt to read channel data not present in the new datalog, throwing exceptions.

brentpicasso:
Various fixes to make analysis more robust:

* Better handling of cases where channels are not available in the selected session; avoid exceptions and needless error messages to the user. #1319,  #1334
* if datalog fails to import, show an error message vs. silently failing.  #1331
* Simplify, centralize and reduce code around logic that selects suggested laps
* Resolve certain cases where we try to add duplicate datapoint columns during import #1351
* Allow channel names starting with a digit. #1330

### DIFF
--- a/autosportlabs/racecapture/views/analysis/analysisview.py
+++ b/autosportlabs/racecapture/views/analysis/analysisview.py
@@ -221,9 +221,10 @@ class AnalysisView(Screen):
     def on_stream_connected(self, instance, new_session_id):
         self.stream_connecting = False
         self._dismiss_popup()
-        session = self._datastore.get_session_by_id(new_session_id)
-        self.ids.sessions_view.append_session(session)
-        self.check_load_suggested_lap(new_session_id)
+        if new_session_id:
+            session = self._datastore.get_session_by_id(new_session_id)
+            self.ids.sessions_view.append_session(session)
+            self.check_load_suggested_lap(new_session_id)
 
     def _set_suggested_channels(self, channels):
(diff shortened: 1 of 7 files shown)


autosportlabs/racecapture/views/analysis/analysisview.py
```diff
@@ -275,14 +266,13 @@ def close_popup(self, *args):
 
     def on_add_session(self, instance, session):
         Logger.info("AnalysisView: on_add_session: {}".format(session))
-        self.check_load_suggested_lap(session.session_id)
```
brentpicasso: redundant call!

autosportlabs/racecapture/views/analysis/linechart.py
```diff
@@ -487,8 +487,8 @@ def get_results(results):
         try:
             self.datastore.get_channel_data(source_ref, ['Interval', 'Distance'] + channels, get_results)
         except Exception as e:
-            alertPopup('Could not load lap', "There was a problem loading the lap due to missing data:\r\n\r\n{}".format(e))
-            raise  e# allow CrashHandler to record the issue
```
brentpicasso: Let the higher level code handle it.

Issue #1347, opened 2017-03-31
Datastore recorder throws exceptions if new channels suddenly appear


brentpicasso:
STR:

* start recording in dashboard
* jump to config, enable a new channel
* write config
* observe datastore throwing exceptions when unknown channel appears

PR #1355, opened 2017-04-06
stop session recording if channel meta changed

brentpicasso:
If new meta is generated (e.g. a new channel was added), need to stop session recording to prevent datastore exceptions.


brentpicasso:
Resolves #1347

Issue #474, opened 2015-10-26
expose alpha filtering value for analog, timer and IMU channels


ryandoherty:
- [X] Analog channels
- [ ] imu channels
- [ ] timer channels

brentpicasso:
Configurable alpha filtering is available in the firmware, need to expose this value in the analog and timer config screens.

PR #1367, opened 2017-04-15
Added Analog Filter alpha to RCP

jlwall:
#474 Partially helps with exposing Alpha for the Analogs

### DIFF
--- a/autosportlabs/racecapture/views/configuration/rcp/analogchannelsview.py
+++ b/autosportlabs/racecapture/views/configuration/rcp/analogchannelsview.py
@@ -72,7 +72,15 @@ class AnalogChannel(BaseChannelView):
                 self.dispatch('on_modified', self.channelConfig)
         except:
             pass
-            
+
+    def on_alpha_map_value(self, instance, value):
+        try:
+            if self.channelConfig:
+                self.channelConfig.alpha = float(value)
+                self.channelConfig.stale = True
+                self.dispatch('on_modified', self.channelConfig)
+        except:
+            pass
                     
     def on_scaling_type_raw(self, instance, value):
         if self.channelConfig and value:
@@ -115,6 +123,7 @@ class AnalogChannel(BaseChannelView):
             check_mapped.active = True
         
         self.ids.linearscaling.text = str(channelConfig.linearScaling)
+        self.ids.alpha.text = str(channelConfig.alpha)
         map_editor = self.ids.map_editor
         map_editor.on_config_changed(channelConfig.scalingMap)
         map_editor.bind(on_map_updated=self.on_map_updated)
--- a/autosportlabs/racecapture/views/configuration/rcp/analogchannelsview.py
+++ b/autosportlabs/racecapture/views/configuration/rcp/analogchannelsview.py
@@ -48,18 +48,18 @@ class AnalogChannelsView(BaseMultiChannelConfigView):
         super(AnalogChannelsView, self).__init__(**kwargs)
         self.channel_title = 'Analog '
         self.accordion_item_height = sp(300)
-        
-            
+
+
     def channel_builder(self, index, max_sample_rate):
         editor = AnalogChannel(id='analog' + str(index), channels=self.channels)
         editor.bind(on_modified=self.on_modified)
         if self.config:
             editor.on_config_updated(self.config.channels[index], max_sample_rate)
         return editor
-            
+
     def get_specific_config(self, rcp_cfg):
         return rcp_cfg.analogConfig
-        
+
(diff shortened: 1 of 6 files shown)


brentpicasso:
Awesome, thank you. This one's been bugging me for a while. :) Will check it out later today.

jlwall:
I think I got some useable for the IMU as well. I know these default @ 0.1    I want you guys to make sure it defaults a correct value when there is no setup to begin with.

Do you think you want limiters to keep the range 0-1?

brentpicasso:
Cool. If you want to go for completeness, the timer channels also have the alpha filter setting too. 

Correct, IMU is 0.1.   There's an entire separate topic to address the confusion with a blank (no read, not loaded config) but that's a topic for a different Issue. 

They should have limiters to keep the range 0-1.

brentpicasso:
oh, also test it using emuphone.sh - this will let you check that the display density you'd typically see on a small form factor device, like a phone

jlwall:
I got all three working, and added limiters, would like a review from the team.
I checked on the emu_phone, and OSX only

brentpicasso:
Hi @jlwall , I was playing with this and realized a few things, and have an idea.

1) users likely won't know what  Alpha means - it's a technical term that's appropriate for the underlying API, but the UI should describe this in a more friendly way. 
2) It's not obvious to the user that a small number represents 'a lot of smoothing' and 1.0 means 'no smoothing'
3) making the setting a little more visual and work better with touch interfaces

Suggestion: use the kivy slider coupled with a text box - with some legends and labels. If the value is directly edited, it updates the slider, and vice-versa. 
image

I think this would go a long way to make it more self explanatory, limit questions and make a more frustration-free experience.  

Thoughts?

autosportlabs/racecapture/views/configuration/rcp/analogchannelsview.py
```diff
@@ -72,26 +72,40 @@ def on_linear_map_value(self, instance, value):
...
-            
-                    
+
+    def on_alpha_map_value(self, instance, value):
+        try:
+            if self.channelConfig:
+                if  float(value) > 1:
+                    value = 1
+                    self.ids.alpha.text = str(value)
+                elif float(value) < 0:
```
brentpicasso: Would need to check this, but I think an alpha of 0 would have the value never update, which would cause confusion for users.  Probably should check this and make the lower range a very small number, like 0.001.

autosportlabs/racecapture/views/configuration/rcp/imuchannelsview.py
```diff
@@ -76,13 +76,31 @@ def enable_view(self, enabled):
...
             self.notify_modified()
 
+    def on_alpha_value(self, instance, value):
+        try:
+            if self.channelConfig:
+                if  float(value) > 1:
+                    value = 1
+                    alpha = kvFind(self, 'rcid', 'alpha')
+                    alpha.text = str(value)
+                elif float(value) < 0:
```
brentpicasso: ditto on min-value comment above

autosportlabs/racecapture/views/configuration/rcp/timerchannelsview.py
```diff
@@ -57,55 +57,71 @@ class TimerModeSpinner(MappedSpinner):
...
             self.channelConfig.stale = True
             self.dispatch('on_modified', self.channelConfig)
-                            
-    def on_config_updated(self, channel_config, max_sample_rate):
-        self.ids.sr.setValue(channel_config.sampleRate, max_sample_rate)
-    
-        self.ids.chan_id.setValue(channel_config)
-        
+
+    def on_config_updated(self, channelConfig, max_sample_rate):
```
brentpicasso: we're migrating towards PEP8 compliance, so variable name should rename with underscores.

autosportlabs/racecapture/views/configuration/rcp/analogchannelsview.py
```diff
@@ -72,26 +72,40 @@ def on_linear_map_value(self, instance, value):
                 self.dispatch('on_modified', self.channelConfig)
         except:
             pass
-            
-                    
+
+    def on_alpha_map_value(self, instance, value):
+        try:
+            if self.channelConfig:
+                if  float(value) > 1:
```
brentpicasso: nit - might want to convert value to float once at the top so you don't do it again on line 85

brentpicasso:
Hi @jlwall , any luck on the suggested changes? Just following up..

brentpicasso:
Hi @jlwall - any progress on this? Would you like us to move the ball forward from here? It's good work so far.  Thanks!

PR #1395, opened 2017-05-23
enhanced channel name validation for input field and datastore

### LINKED ISSUE #1370: limit channel names to alphanumeric characters
Channel names should be limited to alphanumeric characters, with no embedded whitespace. 

* prevent invalid characters
* strip out embedded whitespace

brentpicasso:
resolves #1370 

Also refactored: 

Move channelnameselectorview.py out of root directory and into package hierarchy;
moved .kv layout from .kv file into code files.


autosportlabs/racecapture/views/configuration/channels/channelsview.py
```diff
@@ -37,70 +37,179 @@
 from valuefield import FloatValueField, IntegerValueField, TextValueField
 from autosportlabs.widgets.scrollcontainer import ScrollContainer
 
-Builder.load_file('autosportlabs/racecapture/views/configuration/channels/channelsview.kv')
+
+class ChannelNameField(TextValueField):
+
+    def insert_text(self, s, from_undo=False):
+        s = ''.join([char for char in s if char.isalnum() or char == ' '])
```
ryandoherty: This looks repeated logic from `_scrub_sql_value`, we should centralize that logic/sanitation if we can.

test/autosportlabs/racecapture/datastore/test_datastore.py
```diff
@@ -414,11 +414,12 @@ def parse_header_meta(headers):
...
             self.ds.delete_session(import_export_id)
 
-
-
-
-
-
-
-
+    def test_scrub_sql_value(self):
```
ryandoherty: Yay tests!

autosportlabs/racecapture/views/configuration/channels/channelsview.py
brentpicasso: It does. I was trying to avoid coupling across modules. Let me see if there's a common area / util module we can put it in.

brentpicasso: So I centralized it to a point for UI components. 

For the datastore, the purpose is similar but has a subtly different purpose, and I didn't want to couple the modules together.  I think it's better overall if datastore maintains it's own logic for scrubbing a sql value.

Issue #1388, opened 2017-05-05
Sentry: OperationalError: no such column: LapTime


brentpicasso:
OperationalError: no such column: LapTime
(57 additional frame(s) were not displayed)
...
  File "kivy/_event.pyx", line 1148, in kivy._event.EventObservers._dispatch (kivy/_event.c:12363)
  File "home/brent/git-projects/RaceCapture_App/.buildozer/android/app/autosportlabs/racecapture/views/analysis/analysisview.py", line 278, in on_add_session
  File "home/brent/git-projects/RaceCapture_App/.buildozer/android/app/autosportlabs/racecapture/views/analysis/analysisview.py", line 241, in check_load_suggested_lap
  File "home/brent/git-projects/RaceCapture_App/.buildozer/android/app/autosportlabs/racecapture/datastore/datastore.py", line 870, in get_channel_min
  File "home/brent/git-projects/RaceCapture_App/.buildozer/android/app/autosportlabs/racecapture/datastore/datastore.py", line 862, in _get_channel_aggregate

Issue #1397, opened 2017-05-24
Analysis fails to load data when lap timing is disabled


brentpicasso:
Analysis features assumes the presence of LapCount and CurrentLap. When those channels are missing, session data can not be viewed. 

When importing or logging data with no laptiming data, we need to show a default 'lap' with all of the data in the session, which allows the data to be viewed.

PR #1398, opened 2017-05-24
Allow analysis to view data when no laptiming is present. 

brentpicasso:
Allows viewing of data if LapCount / CurrentLap channels are missing. 

Step to test:

* delete datastore
* disable lap timing in Setup
* write config to unit
* go to dashboard, record some data
* go to analysis, select newly recorded session for viewing

Should be able to see one 'lap' and be able to view data.

### DIFF
--- a/autosportlabs/racecapture/datastore/datastore.py
+++ b/autosportlabs/racecapture/datastore/datastore.py
@@ -28,6 +28,9 @@ import datetime
 from kivy.logger import Logger
 from collections import OrderedDict
 
+class InvalidChannelException(Exception):
+    pass
+
 class DatastoreException(Exception):
     pass
 
@@ -298,7 +301,6 @@ class DataStore(object):
     def __init__(self, databus=None):
         self._channels = []
         self._isopen = False
-        self.datalog_channels = {}
         self.datalogchanneltypes = {}
         self._ending_datalog_id = 0
         self._conn = None
@@ -353,6 +355,9 @@ class DataStore(object):
     @property
     def is_open(self):
         return self._isopen
+
+    def channel_exists(self, channel):
+        return channel in [x.name for x in self._channels]
 
     @property
     def channel_list(self):
@@ -870,6 +875,9 @@ class DataStore(object):
         if sessions is not None:
             params = params + sessions
 
+        if not self.channel_exists(channel):
+            raise InvalidChannelException()
+
         base_sql = "SELECT {}({}) {} from datapoint {} {};".format(aggregate, _scrub_sql_value(channel),
                                                                  self._extra_channels(extra_channels),
                                                                  self._session_select_clause(sessions),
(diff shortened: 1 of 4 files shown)


autosportlabs/racecapture/datastore/datastore.py
```diff
@@ -298,7 +301,6 @@ class DataStore(object):
     def __init__(self, databus=None):
         self._channels = []
         self._isopen = False
-        self.datalog_channels = {}
```
brentpicasso: dead code!

brentpicasso:
resolves #1388 #1397

autosportlabs/racecapture/datastore/datastore.py
```diff
@@ -1084,6 +1117,7 @@ def get_laps(self, session_id):
         for lap in laps:
             laps_dict[lap.lap] = lap
 
+        print('laps dict ' + str(laps_dict))
```
ryandoherty: Debug code?

autosportlabs/racecapture/datastore/datastore.py
```diff
@@ -1040,6 +1048,24 @@ def get_sessions(self):
...
+        :type Boolean
+        '''
+        if not (self.channel_exists('CurrentLap') and self.channel_exists('LapCount')):
+            return False
+
+        c = self._conn.cursor()
+        for row in c.execute('''SELECT s.session_id, d.LapCount, d.CurrentLap FROM sample s, 
+                             datapoint d WHERE s.session_id = ? AND d.sample_id = s.id LIMIT 1;''',
+                                (session_id,)):
+            return row[1] is not None and row[2] is not None
```
ryandoherty: Why not just do a select where d.LapCount and d.CurrentLap is not null? Also, you're going to iterate over all session data for this, which will be very slow on mobile devices with lots of data. I'd try to pre-compute this information while it's being recorded instead. Pre-compute as much as possible, I've seen the analysis view hang a lot when loading sessions on slower tablets (Kindle Fire).

autosportlabs/racecapture/datastore/datastore.py
brentpicasso: auugh!

autosportlabs/racecapture/datastore/datastore.py
brentpicasso: That would work too. I assumed the query would always be fast because there's no sorting and I'm using LIMIT 1.

brentpicasso: I applied timing to that query and it's coming back in 0.1ms where the main query takes 21.7 ms when loading laps from sonoma. 

 [Datastore   ] session_has_laps function took 0.10085105896 ms
[Datastore   ] _query_channel_data function took 21.7950344086 ms

Because the query is not sorted and has LIMIT 1 it appears to be O(1) complexity.

I think a "select where d.LapCount and d.CurrentLap is not null"  might do a table scan even if we do a LIMIT 1, because it could cause it to scan the entire data segment looking for a not null value. 

What I'm doing here is just taking the first row I can, and checking if the columns in question are null.  The rationale is, if the session was imported with LapTime + Current Lap channels present, none of the values in those columns will be null, guaranteed by how the data is populated into datapoint during import.

PR #1399, opened 2017-05-28
Session recording in a background thread

### LINKED ISSUE #1094: SessionRecorder: performance testing
Since SessionRecorder is doing the SQL inserts in the UI thread, we should do some performance testing to see what conditions may cause the app to visually lag.

brentpicasso:
Record session data in a background thread to prevent impacting UI responsiveness.  Resolves #1094

### DIFF
--- a/autosportlabs/racecapture/data/sessionrecorder.py
+++ b/autosportlabs/racecapture/data/sessionrecorder.py
@@ -28,16 +28,20 @@ from autosportlabs.racecapture.geo.geopoint import GeoPoint
 from autosportlabs.util.threadutil import safe_thread_exit
 import copy
 import Queue
-from Queue import Empty
+from Queue import Empty, Full
 
 
-class SessionRecorder(EventDispatcher) :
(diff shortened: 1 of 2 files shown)


autosportlabs/racecapture/data/sessionrecorder.py
```diff
@@ -218,8 +270,16 @@ def _on_sample(self, sample):
         :param sample:
         :return:
         """
-        if self.recording:
-            self._save_sample(sample)
+        if not self.recording:
+            return
+        try:
+            self._sample_queue.put_nowait(copy.deepcopy(sample))
```
ddimensia: Is there a risk that the sample is going to be modified or is this just being defensive?

brentpicasso: correct, the sample instance passed in will get modified by the thread running the databus.

brentpicasso:
agreed; we'd have to put some mutexes in there if we get to that point.  Thanks for the review!

PR #1405, opened 2017-06-03
remember track map selections. resolves #1002

### LINKED ISSUE #1002: automatic track map selection occurs even if user selects a specific map
Scenario: a use selects a specific track map is selected to match a venue configuration (e.g. NJMP lightning). 

If all tracks are de-delected and then selected again, the default trackmap is selected again.

brentpicasso:
Analysis tries to automatically pick a nearby track map, but can pick the wrong one for the current data. Example; sonoma.log will pick the sonoma kart track instead of the main track. 

if you go in and override the track, and then unselect and re-select laps, then it will snap back and auto-detect the wrong track. 

The changes in this PR will remember the selected track and prefer that selection the next time it tries to auto-detect a nearby track. 

Sample logs to test with here:
https://www.dropbox.com/sh/o6u4kil4xdv4udb/AAAoOdszJSWQ9--WqUYz5zkya/sample%20logs?dl=0

Sonoma and Tsukuba logs are both good ones to test with.

Resolves #1002

### DIFF
--- a/autosportlabs/racecapture/views/analysis/analysismap.py
+++ b/autosportlabs/racecapture/views/analysis/analysismap.py
@@ -131,6 +131,7 @@ class AnalysisMap(AnalysisWidget):
     SCROLL_FACTOR = 0.15
     track_manager = ObjectProperty(None)
     datastore = ObjectProperty(None)
+    settings = ObjectProperty(None)
 
     def __init__(self, **kwargs):
         super(AnalysisMap, self).__init__(**kwargs)
@@ -146,6 +147,16 @@ class AnalysisMap(AnalysisWidget):
         Window.bind(on_motion=self.on_motion)
 
 
+    def on_settings(self, instance, value):
+        #initialize our preferences
+        value.userPrefs.init_pref_section('analysis_map')
+        
+    def get_pref_track_selections(self):
+        return self.settings.userPrefs.get_pref_list('analysis_map', 'track_selections')
+        
+    def set_pref_track_selections(self, selections):
+        self.settings.userPrefs.set_pref_list('analysis_map', 'track_selections', selections)
+        
     def refresh_view(self):
         """
         Refresh the current view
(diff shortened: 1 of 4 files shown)


autosportlabs/racecapture/views/analysis/analysismap.py
```diff
@@ -200,8 +211,27 @@ def _update_legend_box_layout(self):
...
-    def _update_trackmap(self, values):
-        track = self.track_manager.get_track_by_id(values.track_id)
+    def _save_selected_trackmap(self, track):
+        # save the selected track in preferences so it is remembered 
+        # for next time a track is selected in the nearby area.
+        saved_tracks = self.get_pref_track_selections()
+        
+        nearby_tracks = self.track_manager.find_nearby_tracks(track.centerpoint)
+        
+        for nearby_track in nearby_tracks:
```
ddimensia: Might be good to have a comment here that explains that the saved tracks can only contain one track within a given area.  The comment below explains what you're doing but not the intent, or at least it wasn't obvious to me the first read through.

ddimensia:
Maybe I'm not understanding the initial bug but the implementation feels like a partial solution.  Using Thunderhill as an example, if I have two sessions, one for Thunderhill West and one for the full 5 mile configuration then if I switch between the two session in analysis it would cause me to have to switch the map selection every time.  What seems to be a better solution would be to associate the track_id with the session data rather than saving it globally (or maybe in addition to).  

I could be mis-understanding the intent and the code but I feel this could lead to a new yet less pervasive bug.  That being said, it would reduce user pain in the case where a kart track is being selected instead of an adjacent road course.

brentpicasso:
Correct, it's an incremental improvement, and there are some cases where it
won't be sufficient.

My sense is to not couple the track map id's with the datastore,  so the
next step to improve it is this issue.
https://github.com/autosportlabs/RaceCapture_App/issues/1406

Brent Picasso
Autosport Labs

autosportlabs/racecapture/views/analysis/analysismap.py
brentpicasso: agreed, will update.

ddimensia:
👍

brentpicasso:
thanks for the review!

Issue #1336, opened 2017-02-21
[iphone layout] Saved session list has ellipsis on time field


mattlqx:
Unable to see the full time on an iPhone 7. Also, I'm not sure what time that's even supposed to be as my last sessions were done at approximately 16:00 PST.

![img_0821](https://cloud.githubusercontent.com/assets/739851/23148397/bc58fb12-f7a2-11e6-84b4-6e5a702d9edb.PNG)

Issue #1092, opened 2016-07-27
SessionRecorder: enhance name and timestamp in saved sessions


brentpicasso:
Name:
- Auto-generate name based on detected track, if possible. 

Timestamp:
have a scheme that's easier to visually recognize when a session occurred.  Riffing off what you see in other apps, like facebook when it shows when something occurred, like a comment. 

Ideas (not all have to be implemented)
- If it happened within an hour, show the session time as (XX minutes ago)
- If it happened today, show session time as (Today at HH:MM:SS)
- If it happened within a week, show session time as (XX days ago)
- If it happened beyond that, just show the date / time.

ryandoherty:
+1, I had this in my todo list later. One big note: timestamps differ per locale, so we need the timestamps we generate use locale-dependent formatting: MM/DD/YY - US, DD/MM/YY - UK, Germany, etc.

ryandoherty:
`strftime` does have locale formatted strings, but not in any customizable form though: https://docs.python.org/2/library/time.html#time.strftime

PR #1407, opened 2017-06-05
enhance usability / readability of session list

### LINKED ISSUE #1375: Session list truncates session timestamp on certain form-factors
It's hard to identify your sessions if you can't see the full time they were created.  This seems to happen mostly on smaller screens, like a phone:

image

brentpicasso:
Resolves the issue of frequent truncation of session name and date time, while making the listing more readable through friendlier descriptions of date time stamps. 

New dialog looks like this:

image

resolves #1375 #1336 #1092

### DIFF
--- a/autosportlabs/racecapture/views/analysis/addstreamview.kv
+++ b/autosportlabs/racecapture/views/analysis/addstreamview.kv
@@ -97,37 +97,31 @@
         id: name
         size: self.texture_size
         font_size: self.height * 0.6
-        size_hint_x: 0.38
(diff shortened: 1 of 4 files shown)


ddimensia:
If the icons are fixed size, which I believe they are, it would be best if the layout specified the icons with fixed widths and allow the name/date to take the rest.

brentpicasso:
Thanks for the review.  I updated the kivy layout for the buttons so they're fixed width, UI looks good on both emu_phone.py and regular launch.

PR #1417, opened 2017-06-10
further fix channel cleansing, permit space and underscore in channel…

### LINKED ISSUE #1416: allow spaces and underscores in channel names, filter out everything else.
Session recording with spaces is causing an error, in addition to session exports. Showing up in 1.10.0. 

Should allow spaces and underscores in channel names, but filter out everything else.

brentpicasso:
… names. resolves #1416


brentpicasso:
self reviewing, relatively minor change.

PR #1455, opened 2017-07-17
Add CAN termination controls in CAN config view.

### LINKED ISSUE #1442: expose CAN termination switches
expose CAN termination switches as provided by firmware

brentpicasso:
+ related refactorings for PEP8 compliance; moved .kv file inside source. 

resolves #1442


autosportlabs/racecapture/config/rcpconfig.py
```diff
@@ -908,23 +908,38 @@ def __init__(self, **kwargs):
...
-        bauds = canCfgJson.get('baud')
+    def fromJson(self, can_cfg_json):
+        self.enabled = True if can_cfg_json.get('en', self.enabled) == 1 else False
+
+        bauds = can_cfg_json.get('baud')
         self.baudRate = []
         for baud in bauds:
             self.baudRate.append(int(baud))
+
+        terms = can_cfg_json.get('term')
```
ddimensia: Wouldn't it be cleaner to just use can_cfg_json.get('term', [0,0]) and drop the None check?  There's also some risk here of a malformed config causing issues, e.g. empty array, etc.  I have a feeling there's a general lack of json validation in the code so I'm not sure this PR is the place to solve that.

brentpicasso: Although it's a little less explicit, it's also more concise. I like it. 

It seems there is at least one approach to json schema validation a'la XML validation, but yes, it's out of scope for this PR: http://json-schema.org/

brentpicasso: Ok, fix applied.

Issue #1322, opened 2016-12-23
Re-work analog mapping editor


brentpicasso:
Analog mapping editor should be a popup window instead of being present directly in the form. 

The popup should allow any values to be entered, and validate at the end, instead of attempting to validate on the fly, which often gets in the way of data input.

Issue #983, opened 2016-05-15
Organize and store CAN mapping presets


brentpicasso:
Ability to retrieve a list of preset CAN mapping configurations to quickly enable support for ECUs, sensors and etc.
- Similar to how trackmaps are managed,
  *\* Pre-load app distribution with a default collection of currently known mappings
  *\* Ability to download updated preset mappings

When a preset is selected, offer the option to replace or append to the current channel mappings.

![image](https://cloud.githubusercontent.com/assets/3267533/15275779/8e30a682-1a89-11e6-8f82-0b7367c2723b.png)

Issue #783, opened 2016-01-03
Analog sensor presets


brentpicasso:
Provide a browseable list of analog sensor pre-sets for well known sensors. 

Once a sensor is selected, it would import the calibration and min/max values.

brentpicasso:
Leverage the same infrastructure as the upcoming CAN presets.

Issue #491, opened 2015-10-26
add offset value for linear mode in analog channel


ryandoherty:

brentpicasso:
implemented in the config, need to expose in UI.

PR #1472, opened 2017-10-16
Presets for CAN mapping, OBDII and Analog channels

### LINKED ISSUE #985: Organize and store OBD-II mapping presets
Similar to https://github.com/autosportlabs/RaceCapture_App/issues/983, have the ability to select from a list of presets (presumably year / make /model vehicle) and auto populate a list of known PIDs, including known OEM specific PIDs.

Similar to trackmap management:
- Pre-load app distribution with a default collection of currently known mappings
- Ability to download updated preset mappings

brentpicasso:
This hooks up presets for the CAN mapping, OBDII and analog channel view. 

resolves #985 #983 #783 

also re-worked look of analog mapping view, and enabled the linear offset field in addition to multiplier so it looks like a linear formula - #491 #1322

Pane of map vs formula vs none updates based on the selected mapping type.

image

image

image


ddimensia:
Just an initial comment on the screen shots above, could be done as a follow-up issue/PR.

It would be nice if all three modes of setting up the analog gauges kept the graph to the right so that it shows what the result would be.  Even though it seems obvious what raw 0-5v will be I think it's better for it to be consistent so users have a visual representation.  This would also allow the UI to have a consistent layout and not jump around.

Also, in the formula case it would be nice if you could put labels below the multiplier and offset text inputs, e.g. (multiplier), (offset).

autosportlabs/racecapture/presets/presetview.py
```diff
@@ -0,0 +1,225 @@
+import kivy
```
ddimensia: Copyright header

autosportlabs/racecapture/views/configuration/rcp/canchannelsview.py
```diff
@@ -8,7 +8,7 @@
 from kivy.uix.popup import Popup
```
ddimensia: Add Copyright declaration

brentpicasso:
You're right, having the chart there all the time would make sense to illustrate the effects of the formula or mapping. I'll see what can be done in the short term.

autosportlabs/racecapture/presets/__init__.py
```diff
@@ -0,0 +1,20 @@
+#
+# Race Capture App
+#
+# Copyright (C) 2014-2016 Autosport Labs
```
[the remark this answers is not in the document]
brentpicasso: will sweep through the main codebase to handle that, in commit on the main branch.

brentpicasso:
Reworked analog mapping view so graph is always visible: 

blob:https://autosportlabs.imgur.com/766d96cc-f0c0-40a7-899e-385fee947c3f

